PR #75, opened 2022-01-27
Feature/index-page

kiwiidb:
Adds a landing page like https://lndhub.io

Still to do:
- Replace Bluewallet logo with Alby logo
- Custom styling
- Add caching so we don't hit the Lightning node on each page load. We can have an expiration of a couple of minutes here.


bumi (changes requested):
looks good. we should `embed` the static files to make sure we still have only one executable. 

and URIs were empty for me / not part of the getInfo response.

main.go
```diff
@@ -142,7 +142,9 @@ func main() {
 	blankController := controllers.NewBlankController(svc)
 	secured.GET("/getbtc", blankController.GetBtc)
 	secured.GET("/getpending", blankController.GetPending)
-	e.GET("/", blankController.Home)
+	e.GET("/", controllers.NewHomeController(svc).Home)
+	e.GET("/qr", controllers.NewHomeController(svc).QR)
```
bumi, line 146: what's a good practice here actually? 
shouldn't this be one home controller instance?

kiwiidb: Yes, fixed. I feel like we can use a single controller for the two routes here.

kiwiidb:
For the URI's, `lncli getinfo` is not showing any URI's for our nodes, neither on regtest or mainnet. Not sure why this is the case.
We could add our LoadBalancer's IP address as a configuration parameter to LNDHub, then we would be able to construct this URI by hand.

main.go
```diff
@@ -31,6 +32,12 @@ import (
 	"github.com/ziflex/lecho/v3"
 )
 
+//go:embed templates/index.html
+var html string
```
bumi, line 36: maybe indexHtml? to be more descriptive that this is specifically the index page?

bumi (approved):
maybe we just add a link to amboss.space with the pub key instead of the URIs?

PR #76, opened 2022-01-27
Fix/error handling

bumi:


kiwiidb (approved):
Verified that this changes the error code to 400.

PR #80, opened 2022-01-31
Avoid using mysql-specific syntax for timestamp

kiwiidb:
Paid invoices won't be picked up when using SQLite without this fix.


kiwiidb:
Duplicate of https://github.com/getAlby/lndhub.go/pull/79

PR #87, opened 2022-02-01
Feature/integration-testing

kiwiidb:
Add integration tests with testify/suite. Moved some structs
outside of controller funcs so we can re-use them in the testing
package. Add CI workflow for running tests on every push.

### DIFF
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -1,0 +1,18 @@
+name: Integration tests
+on: push
+jobs:
+  test:
+      strategy:
+        matrix:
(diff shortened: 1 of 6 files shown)


bumi (approved):
Sadly I don't have experience in testing Go apps.

The test running automatically runs all Test* functions or how does this work here in Go?

kiwiidb:
Test files should have the suffix `_test.go`.
Tests are funcs that start with `Test` and have a *testing.T as a param.
The command in the workflow will always run all tests in the integration_tests directory and print out the coverage.
We can add more complex behaviour later, like a bot that comments on a PR if there is a failed test or something.

PR #90, opened 2022-02-04
Feature/wrap-lnd

kiwiidb:
Wrapping the Lightning Client allows us to plug in different Lightning back-ends.
I've opted to keep using the data structures defined by the lnrpc package.
I've also needed to wrap the Subscriber, because other subscriptions don't need an entire grpc subscriber.

### DIFF
--- a/lib/service/invoicesubscription.go
+++ b/lib/service/invoicesubscription.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/getAlby/lndhub.go/db/models"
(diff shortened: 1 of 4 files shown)


bumi:
great. shouldn't we then call it `ln.LightningClientWrapper` instead of `lnd` ?

kiwiidb:
Yes. But then we rename the package which might need some more refactoring.

bumi (approved):

PR #91, opened 2022-02-04
Feature/integration-testing-payments

kiwiidb:


integration_tests/create_test.go
```diff
@@ -63,39 +66,77 @@ func TestCreateUserTestSuite(t *testing.T) {
...
+		logger.Fatalf("Failed to decode IdentityPubkey: %v", err)
+	}
+	identityPubKey, err := btcec.ParsePubKey(hexPubkey[:], btcec.S256())
+	if err != nil {
+		logger.Fatalf("Failed to parse node IdentityPubkey: %v", err)
+	}
+	svc.IdentityPubkey = identityPubKey
+	logins = []controllers.CreateUserResponseBody{}
+	tokens = []string{}
+	for i := 0; i < usersToCreate; i++ {
```
bumi, line 126: I feel like this function now does a lot of things. 
can't this user preparation be some function on some test service that we can split this up: first init and then create some users? 
We might not only need to create users but also invoices and other things in the future?

kiwiidb: It does do a lot of things, but creating users and fetching tokens is tested seperately.

Adding invoices can be done through [this function](https://github.com/getAlby/lndhub.go/blob/feature/integration-testing-payments/integration_tests/internal_payment_test.go#L76), which is used by multiple tests.

Also, you can't define a function on a receiver from another package. We can split it up but I think it's not needed as you'd want users in practically all of the tests.

kiwiidb: I've split up the function.

bumi (approved):
lgtm! (when tests are green :D )
thanks

integration_tests/create_test.go
bumi: 👍

Issue #55, opened 2022-01-20
Review context usage


bumi:
we mostly use `context.TODO()` is this correct? `TODO` sounds to me like this should not be used? 
what can we improve here?

skosito:
Hey @bumi this seems like a good first issue, is it ok that I pick it up? It seems that I don't have permission to assign myself, could you please do it?

bumi:
sounds great!, please do. thanks!

PR #95, opened 2022-02-08
Remove context todo

skosito:
Reference issue: https://github.com/getAlby/lndhub.go/issues/55

There are 2 methods for creating contexts, context.Background() and context.TODO(). They both return empty context, only difference is in name: TODO should be used on places where we are not sure about if we need context, or context behavior.
However, it seems that repo currently passes context to external package db methods, and I assume that is because this context can be used to timeout or cancel db operations. Using TODO doesn't make much sense here since we are just passing context to these methods, but we know the usage, so we should pass in parent context.

According to this https://www.ardanlabs.com/blog/2019/09/context-package-semantics-in-go.html
Incoming requests to a server should create a Context, but because http.Request already has context we can use that one as a top level, which defaults to empty context.Background.
I did this in 2 steps:
- located context.TODO in service methods and replaced it with ctx passed in as first argument (added ctx as first argument on places where it was not there already)
- pass in request context to service methods from controller handlers

Before changes, I noticed some of services methods already had ctx as first argument, some did not. Some service methods used ctx passed in to method, but some created TODO, some used combination. This PR also makes using context a bit more uniform across the app.

### DIFF
--- a/controllers/addinvoice.ctrl.go
+++ b/controllers/addinvoice.ctrl.go
@@ -45,7 +45,7 @@ func (controller *AddInvoiceController) AddInvoice(c echo.Context) error {
 	}
 	c.Logger().Infof("Adding invoice: user_id=%v memo=%s value=%v description_hash=%s", userID, body.Memo, amount, body.DescriptionHash)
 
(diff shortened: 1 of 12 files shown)


kiwiidb (approved):
LGTM!

Issue #63, opened 2022-01-21
Move invoice states to constants


bumi:
Currently we use the strings in the different places in the code. 
We should use some constants for that

skosito:
Hey @bumi , this one is next unassigned issue in list. Could you please assign it to me, or give me permission to do it? Thanks!

PR #98, opened 2022-02-09
Add constants

skosito:
Referenced issue: https://github.com/getAlby/lndhub.go/issues/63

Created common package with constants for account type, invoice type and invoice state.

### DIFF
--- a/common/globals.go
+++ b/common/globals.go
@@ -1,0 +1,12 @@
+package common
+
+const (
+	InvoiceTypeOutgoing = "outgoing"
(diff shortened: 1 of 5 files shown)


kiwiidb (approved):
LGTM, checks are passing, good to go.

bumi (approved):
lgtm
was wondering about the `common` namespace... or if something like `common.constants.InvoiceStateOpen` (however that is done in Go) is a bit more descriptive. 
but lgtm!

PR #100, opened 2022-02-09
Custom usernames and password when creating new users

skosito:
Just a simple user creation with provided login and password. Also added test for this case.

Should we add some validation here?

### DIFF
--- a/controllers/create.ctrl.go
+++ b/controllers/create.ctrl.go
@@ -25,6 +25,8 @@ type CreateUserResponseBody struct {
 func (controller *CreateUserController) CreateUser(c echo.Context) error {
 	// optional parameters that we currently do not use
 	type RequestBody struct {
+		Login       string `json:"login"`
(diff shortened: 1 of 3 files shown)


kiwiidb:
I don't know if we actually need this, as it is not part of the LNDHub API.
If users create their wallet through Bluewallet, this won't be used.
If users create their wallet through the browser extension or through the Alby web app, we can implement this.

But do we really want to let users do this? We'd need to check if the password is strong enough. If they create the password inside the extension, I don't know how well password manager extensions will be able to interact with this? And will users actually be interested in this? After all, the best protection against a lost account would be that they simply import the browser wallet into Bluewallet, so they will always have a recovery mechanism in case of losing one device.

kiwiidb (changes requested):
Overall looks good and I don't have any objections given that the user-provided password is opt-in. Just need to export the structs and then it's ok.

PR #103, opened 2022-02-11
Get incoming/outgoing invoices tests

skosito:
- Added tests for gettxs ctrl.
- Moved some common code to util file
- Left comments on places for discussion points

### DIFF
--- a/integration_tests/gettxs_test.go
+++ b/integration_tests/gettxs_test.go
@@ -1,0 +1,142 @@
+package integration_tests
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/getAlby/lndhub.go/controllers"
+	"github.com/getAlby/lndhub.go/lib"
+	"github.com/getAlby/lndhub.go/lib/responses"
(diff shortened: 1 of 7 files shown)


integration_tests/gettxs_test.go
```diff
@@ -0,0 +1,165 @@
...
+		log.Fatalf("Error initializing test service: %v", err)
+	}
+	users, userTokens, err := createUsers(svc, 1)
+	if err != nil {
+		log.Fatalf("Error creating test users %v", err)
+	}
+	// Subscribe to LND invoice updates in the background
+	// store cancel func to be called in tear down suite
+	ctx, cancel := context.WithCancel(context.Background())
+	suite.invoiceUpdateSubCancelFn = cancel
```
skosito, line 79: is there a better way to close this subscription? this test suite and following both subscribe, i was getting weird issues until i noticed we need to close this in tear down suite
if this is ok way, i would do the same for other tests

integration_tests/util.go
```diff
@@ -0,0 +1,162 @@
+package integration_tests
```
skosito, line 1: just moved duplicated code i noticed in tests to separate file

kiwiidb (approved):
Implemented my comments, I'll merge now.

Issue #96, opened 2022-02-09
Remove unnecessary dependencies


kiwiidb:
I believe the following dependencies should be removed:
 
- https://github.com/lightningnetwork/lnd/zpay32. This dependency is currently used to decode bolt11 invoices. I think it's a better idea to use the Lightning node back-end to decode invoices that returns a simple struct defined in the lnd package.
- https://github.com/btcsuite/btcd/btcec. This dependency is used to get the back-end node's public key. I think it's simpler if we just stored this as a string.
- https://github.com/btcsuite/btcd/chaincfg. Dependency of the zpay32 package.

I think these changes should help us when we implement a second node back-end.

bumi:
so much less code now also :) 👍

PR #105, opened 2022-02-15
Chore/remove-unnecesary-deps

kiwiidb:
Resolves https://github.com/getAlby/lndhub.go/issues/96

### DIFF
--- a/lib/service/invoices.go
+++ b/lib/service/invoices.go
@@ -6,15 +6,12 @@ import (
 	"encoding/hex"
 	"errors"
 	"math/rand"
-	"strings"
 	"time"
 
-	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/getAlby/lndhub.go/common"
 	"github.com/getAlby/lndhub.go/db/models"
 	"github.com/labstack/gommon/random"
 	"github.com/lightningnetwork/lnd/lnrpc"
-	"github.com/lightningnetwork/lnd/zpay32"
 	"github.com/uptrace/bun"
 	"github.com/uptrace/bun/schema"
 )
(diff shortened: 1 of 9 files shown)


lnd/lnd.go
```diff
@@ -15,6 +15,10 @@ import (
 	"gopkg.in/macaroon.v2"
 )
 
+const (
+	MSAT_PER_SAT = 1000
```
bumi, line 19: where is this used?

lib/service/invoices.go
```diff
@@ -223,32 +215,19 @@ func (svc *LndhubService) PayInvoice(ctx context.Context, invoice *models.Invoic
...
-		ph := *decodedInvoice.PaymentHash
-		invoice.RHash = hex.EncodeToString(ph[:])
-	}
-	if decodedInvoice.MilliSat != nil {
-		msat := decodedInvoice.MilliSat
-		invoice.Amount = int64(msat.ToSatoshis())
+		DestinationPubkeyHex: decodedInvoice.Destination,
+		DescriptionHash:      decodedInvoice.DescriptionHash,
+		Memo:                 decodedInvoice.Description,
+		ExpiresAt:            bun.NullTime{Time: time.Unix(decodedInvoice.Timestamp, 0).Add(time.Duration(decodedInvoice.Expiry))},
```
bumi, line 230: is a unit needed for that `time.Duration`? or is that default seconds?

bumi (approved):
utACK

lnd/lnd.go
kiwiidb: Hmm, must have slipped in from my C-Lightning work, not used anywhere atm, deleted.

Issue #97, opened 2022-02-09
Handling of in transit payments


kiwiidb:
Currently, we just call "SendpaymentSync" and depending on the result

- mark the payment as completed and complete the database transaction
- rollback the database transaction in case of an error

However, it might be possible that the call returns an "in transit" state of the payment. In this case, it is possible that the transaction is completed later. We should investigate if this is possible and if so, account for these kind of payments.

Another possible bug might be when a second request to make a payment comes in while the first is still pending. Does the database lock prevent this from happening, or is this possible? If it is possible, the pending amount should not be accounted in the user's total balance.

A possible solution might be to commit pending payments in the database, and update them once we are sure they either succeeded or failed. At startup time, we can check the status of all pending payments in the database against the Lightning node (reconciliation).

kiwiidb:
Cfr. LNDHub does this asynchronously: https://github.com/BlueWallet/LndHub/blob/master/controllers/api.js#L305

bumi:
True, we probably should not rollback but directly commit each DB transaction. This makes sure the funds are moved from the current account to the outgoing account. 
On error we can then move the funds back from the outgoing account to the current account and mark the invoice as errored. 

but it might be hard to get an "on error" event and only get a settled event? do you know that?

Issue #106, opened 2022-02-15
Save error status on outgoing invoice if the payment fails


bumi:
For outgoing invoice payments we store an invoice but we currently can not differentiate between invoices that did not get paid because of not enough balance or some real payment error. 

We should store the payment error and mark the invoice as failed if the payment fails.

https://github.com/getAlby/lndhub.go/blob/main/controllers/payinvoice.ctrl.go#L89
https://github.com/getAlby/lndhub.go/blob/main/controllers/payinvoice.ctrl.go#L82

skosito:
Should we close this one as well @kiwiidb ?

PR #107, opened 2022-02-16
Fix/in-transit

kiwiidb:
Resolves https://github.com/getAlby/lndhub.go/issues/97

This PR will handle in transit payments, with the exceptions of some edge-cases listed below.

It will do this by always committing the outgoing transaction before trying to pay, and then possibly refunding the user in case of a payment failure. In case of payment failure, the outgoing invoice should also be updated with the failure reason.

The edge cases that should be handled in a later PR:
- LNDhub restarts while a payment is in transit. We can handle this by checking pending payments at startup and also handling any failed payment here.
- LND restarts while payment is pending, or there is a network failure while the payment is pending. This is a more difficult scenario to handle, and I am not sure how we should do this.


lib/service/invoices.go
```diff
@@ -159,35 +158,29 @@ func (svc *LndhubService) PayInvoice(ctx context.Context, invoice *models.Invoic
...
-		paymentResponse, err = svc.SendInternalPayment(ctx, &tx, invoice)
+		paymentResponse, err = svc.SendInternalPayment(context.Background(), invoice)
 		if err != nil {
-			tx.Rollback()
+			svc.HandleFailedPayment(context.Background(), invoice, entry, err)
 			return nil, err
 		}
 	} else {
-		paymentResponse, err = svc.SendPaymentSync(ctx, &tx, invoice)
+		paymentResponse, err = svc.SendPaymentSync(context.Background(), invoice)
```
skosito, line 181: here context.background was missing, also on handleSuccessfulPayment, added it on both places
does it make sense to use same context for couple of these or new one for all?

kiwiidb: I think it doesn't really matter, we can use a new one to make it more clear. It's an empty context anyway.

lib/service/invoices.go
```diff
@@ -196,23 +189,40 @@ func (svc *LndhubService) PayInvoice(ctx context.Context, invoice *models.Invoic
...
 	if err != nil {
-		tx.Rollback()
-		return nil, err
+		// TODO: error logging
+		return err
 	}
 
-	// Commit the DB transaction. Done, everything worked
-	err = tx.Commit()
+	// TODO: maybe save errors on the invoice?
```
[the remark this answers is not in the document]
kiwiidb: Yeah, I think you can add a field for the error here, doesn't need to be a seperate PR.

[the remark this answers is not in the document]
kiwiidb: The relevant issue is https://github.com/getAlby/lndhub.go/issues/106.

kiwiidb (commented):
Nice work! Could you add another test for the following:

- A mock LND wrapper that has an internal channel. When SendPaymentSync is called, it waits for a message on this channel, if it receives the message it will fail the payment.
- Make an external payment, but give the request a context that is canceled after some time.
- Check the balance and assert the payment value is deducted
- Send a message to the channel so the payment fails
- Check the balance again and assert that nothing has been deducted.

I think if we have this test then the PR is ok to merge.

lib/service/invoices.go
```diff
@@ -196,23 +189,46 @@ func (svc *LndhubService) PayInvoice(ctx context.Context, invoice *models.Invoic
...
+		CreditAccountID: entryToRevert.DebitAccountID,
+		DebitAccountID:  entryToRevert.CreditAccountID,
+		Amount:          invoice.Amount,
+	}
+	_, err := svc.DB.NewInsert().Model(&entry).Exec(ctx)
 	if err != nil {
-		tx.Rollback()
-		return nil, err
+		svc.Logger.Errorf("Could not insert transaction entry user_id:%v invoice_id:%v", invoice.UserID, invoice.ID)
+		return err
```
bumi, line 208: those errors are still sent to Sentry? or should we do that explicitly here? 
We say above we run those in a Background context to ignore the request context. 

I just think we have to make sure that we have notifications for such exceptions. This one would require some admin to do something.

lib/service/invoices.go
```diff
@@ -196,23 +189,46 @@ func (svc *LndhubService) PayInvoice(ctx context.Context, invoice *models.Invoic
...
-	// Commit the DB transaction. Done, everything worked
-	err = tx.Commit()
+	invoice.State = common.InvoiceStateError
+	if failedPaymentError != nil {
+		invoice.ErrorMessage = failedPaymentError.Error()
+	}
 
+	_, err = svc.DB.NewUpdate().Model(invoice).WherePK().Exec(ctx)
 	if err != nil {
-		return nil, err
```
bumi, line 212: same here: sentry notification?

lib/service/invoices.go
```diff
@@ -196,23 +189,46 @@ func (svc *LndhubService) PayInvoice(ctx context.Context, invoice *models.Invoic
...
+}
 
-	return &paymentResponse, err
+func (svc *LndhubService) HandleSuccessfulPayment(ctx context.Context, invoice *models.Invoice) error {
+	invoice.State = common.InvoiceStateSettled
+	invoice.SettledAt = schema.NullTime{Time: time.Now()}
+
+	_, err := svc.DB.NewUpdate().Model(invoice).WherePK().Exec(ctx)
+	if err != nil {
+		svc.Logger.Errorf("Could not update sucessful payment invoice user_id:%v invoice_id:%v", invoice.UserID, invoice.ID)
```
bumi, line 229: sentry notification?

kiwiidb:
Looks good to me, but @bumi must approve because I started the PR.

bumi (approved):
looks good!

PR #115, opened 2022-02-22
Use sql for init migration

skosito:
The way I did this is I only used init migration and perform pg_dump using pg admin. I left same timestamp.
Then I removed not needed statements, eg setting table owners to "user", and some flags pg_dump is setting. Examples:

```

SET statement_timeout = 0;
SET lock_timeout = 0;
SET idle_in_transaction_session_timeout = 0;
SET client_encoding = 'UTF8';
SET standard_conforming_strings = on;
SELECT pg_catalog.set_config('search_path', '', false);
SET check_function_bodies = false;
SET xmloption = content;
SET client_min_messages = warning;
SET row_security = off;

SET default_tablespace = '';

SET default_table_access_method = heap;


ALTER TABLE public.accounts OWNER TO "user";
```

Then I migrated using new sql migration and compared 2 pg_dumps, they were identical, checked using diff tool.
Integration tests are passing, and manually running app locally was without errors.

Open question:
- Do we need `down` migration for each `up` migration? 
In case db rollback is needed, that is not possible without `down` migration, code for rollback contains this:

```
	if !cfg.nop && migration.Down != nil {
			if err := migration.Down(ctx, m.db); err != nil {
				return nil, err
			}
		}
```
Even though I can not find this on bun docs, I believe writing `down` migrations should be good practice.


bumi (changes requested):
I was wondering about the sequences, if that can not be easier done with a `SERIAL PRIMARY KEY` in the `CREATE TABLE`? 
but I guess that's fine. 

Thanks for checking the diff! good idea!

We also miss some indexes, should we add those in a new migration? (this is probably the cleanest?)

bumi:
Oh and I think we do not need a `down` for this initial migration. Generally it's a good practice, but this one actually an not be migrated down.

kiwiidb (approved):
LGTM, but I'm not really knowledgeable about SQL stuff, so I'll leave it to @bumi

skosito:
> LGTM, but I'm not really knowledgeable about SQL stuff, so I'll leave it to @bumi

thank you @kiwiidb , let's wait for @bumi 's review before merge then :)

db/migrations/20220115171101_init.up.sql
```diff
@@ -0,0 +1,72 @@
...
+    error_message character varying,
+    add_index bigint,
+    created_at timestamp with time zone DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    expires_at timestamp with time zone,
+    updated_at timestamp with time zone,
+    settled_at timestamp with time zone,
+    CONSTRAINT fk_user
+        FOREIGN KEY(user_id) 
+        REFERENCES users(id)
+        ON DELETE CASCADE
```
bumi, line 34: just to double be sure: this deletes the invoice of a user when the user is deleted. but NOT the other way around, correct?

bumi (approved):
utACK

db/migrations/20220115171101_init.up.sql
skosito: Yes, child table entry deleted when parent table entry is deleted

PR #116, opened 2022-02-24
Refresh token fix

skosito:
I did following changes:
- set different expiry for access and refresh tokens in config (2 and 7 days)
- return responses.BadAuthError in case of login error, i think this is enough to fix issue 109, because in case we return response, generic http error handler won't be hit, and generic sentry capture is skipped - please correct me if i am wrong here, it works that way locally, and according to echo docs
- extended test for login with only refresh token
- pull user id from refresh token claim, added util method for this, which parse token with claims, check validity, check isRefresh claim and extract userId claim
- get user by that user id and continue with same flow


kiwiidb (changes requested):
I would like to have a unit test that tries to login with an invalid refresh token, either an unparsable JWT, a JWT with an invalid signature, an expired JWT, or all of these.

Other than that, this LGTM.

skosito:
> I would like to have a unit test that tries to login with an invalid refresh token, either an unparsable JWT, a JWT with an invalid signature, an expired JWT, or all of these.
> 
> Other than that, this LGTM.

Thanks for review @kiwiidb . I added failing tests for following cases:
- expired token
- invalid secret
- invalid user id
- access token
- non parseable token

Please re-check.

kiwiidb (approved):
LGTM

PR #117, opened 2022-02-25
Fee handling

skosito:
Add fixed fee according to issue description and extended tests.

Tests are failing currently when using lnd1 and lnd2 with `rpc error: code = Unknown desc = server is still in the process of starting`.

To add and execute tests I had to configure local polar network and execute using that one, maybe good candidate for future is to have more dynamic configuration for tests, so switch can be easier.


kiwiidb (changes requested):
I have added a commit that fetches the fee amount from the Lightning transaction. It might not have been clear from the issue, but this is what we actually want to do. However, this broke the integration tests ofcourse, sorry for that 😅.

We need to have integration tests that test the fees, so I will add support for a third node, one that doesn't have a direct channel to `simnet-lnd-1`, so we can use that node to generate an invoice. The payment will then be routed through `simnet-lnd-2`, which will charge a fee. Then we can fetch the LN payment, check the fee and check if it corresponds to the fee that we charged the user.

There is another issue, and it is that we currently don't check beforehand if the user can afford the fee. This could potentially lead to a user having a negative balance. Although, I think this issue should be handled in a seperate PR. Bluewallet  does this by checking that the user can afford the payment + 1% fee, before doing the payment. We could do something similar but we have to watch out for UX pitfalls. But as I said, that should be a different PR.

lib/service/invoices.go
```diff
@@ -190,7 +190,8 @@ func (svc *LndhubService) PayInvoice(ctx context.Context, invoice *models.Invoic
 
 	// The payment was successful.
```
bumi, line 191:
```suggestion
	// The payment was successful.
	// these changes to the invoice are persisted in the `HandleSuccessfulPayment` function
```

To make it clear that the saving happens in the function.

lib/service/invoices.go
```diff
@@ -223,11 +224,33 @@ func (svc *LndhubService) HandleFailedPayment(ctx context.Context, invoice *mode
...
+		sentry.CaptureException(err)
+		svc.Logger.Errorf("Could not insert fee transaction entry user_id:%v invoice_id:%v", invoice.UserID, invoice.ID)
+		return err
+	}
+
 	invoice.State = common.InvoiceStateSettled
 	invoice.SettledAt = schema.NullTime{Time: time.Now()}
 
-	_, err := svc.DB.NewUpdate().Model(invoice).WherePK().Exec(ctx)
+	_, err = svc.DB.NewUpdate().Model(invoice).WherePK().Exec(ctx)
```
bumi, line 253: if we fail to save the transaction entry the invoice will not be update. 
I think it makes sense to first save the invoice and then save the fee transaction entry. 
This way we do not loose data (preimage and that the invoice was actually paid) on the invoice if we somehow fail to save the transaction entry (unlikely though)

kiwiidb:
I have added lnd-3 credentials.

db/models/invoice.go
```diff
@@ -14,6 +14,7 @@ type Invoice struct {
 	UserID               int64        `json:"user_id" validate:"required"`
 	User                 *User        `bun:"rel:belongs-to,join:user_id=id"`
 	Amount               int64        `json:"amount" validate:"gte=0"`
+	Fee                  int64        `json:"fee" validate:"gte=0"`
```
bumi, line 17: can this be zero here? I think we save the invoice before we know the fee amount. ?

kiwiidb: It can be zero if paid to direct channel partner or internal.
I added it so it could be passed on to the `HandleSuccessfulPayment` func.

bumi:
I thought the ideas was to use that fixed fee because we do not know the fee amount before the actual transaction. 

With the 1% reserve the user can not fully empty their account easily - which for me is a no-go. 
IMO when lndhub can charge an additional transaction fee then it can also afford a potential higher fee when emptying the account.

Because of the DB constraints the account can not go negative I think. So saving the transaction entry would fail.

db/models/invoice.go
bumi: sorry I meant `NULL`

kiwiidb: Oh, like that. Yes I think we could make it nullable to differentiate between a not-yet-known fee and a 0 fee.

kiwiidb:
Which fixed fee do you mean? If we don't add any check beforehand this could easily be exploited. Construct route so you receive part of the routing fee, pay yourself, you won't be (fully) charged for the fee, refund wallet, repeat.

We could add an endpoint to empty your wallet to a specific Lightning Address, using an algorithm that pays out in a series of payments, where the last transaction will be sufficiently small so that you can end up at 0 sats.

skosito:
Hey @bumi @kiwiidb , thanks for review, I fixed these smaller comments, but it seems from the comments it is still not decided if we should go with fixed fees or not. Going to wait for that decision to be confirmed before fixing that part, please let me know.

bumi:
I actually don't think a special endpoint is an option. this is bad UX and also not implemented in wallets. 

Even if this is a technical problem the risk here is to loose sats when doing that last transaction. And this risk should be rather easy to calculate. We can set max fee amounts on outgoing transactions to catch potential fraud that would be too expensive. 

I'd rather loose on one transaction the transaction fee (for which I can set a maximum) than annoying users by telling them they can not get their funds and not having a clear answer on how much they can withdraw. 
Surely this is a problem of how lightning works but we have to somehow deal with that.

kiwiidb:
OK, no endpoint, but it can still be solved at the client side by doing multiple transactions.

> We can set max fee amounts

I don't think this will work:

- User figures out what the max fee for a transaction is. Fixed or percentage of the amount, doesn't matter.
- User constructs an invoice that pays the max fee (or `max fee - fee charged by previous hop(s)`) to themselves.
- Refund wallet, for which the user will pay less fees than the outgoing transaction.
- Repeat for infinite money.

So in fact, not doing a fee reserve check beforehand is the same security as not charging fees at all. Unless I'm missing something here.

kiwiidb:
> A user could create many accounts but we should notice that anyway.

How are we going to notice this if it is done in a distributed fashion?

db/migrations/20220120000700_add_constraints.up.go
```diff
@@ -60,7 +71,7 @@ func init() {
 					WHERE account_ledgers.account_id = NEW.debit_account_id;
 
 					-- IF the account would go negative raise an exception
-					IF sum < 0 AND debit_account_type != 'incoming'
```
skosito, line 63: at this point we know that debit_account_type can not be incoming?

db/migrations/20220120000700_add_constraints.up.go
```diff
@@ -48,8 +49,18 @@ func init() {
 					--   This can happen when two transactions try to access the same account
 					FOR UPDATE NOWAIT;
 
-					-- If it is an incoming account return; otherwise check the balance
-					IF debit_account_type IS NULL
+					-- check if credit_account type is fees, if it's fees we don't check for negative balance constraint
+					SELECT INTO credit_account_type type
```
skosito, line 53: just did the same as above, checked if credit account type is fees, if it is don't check for sum constraint, is this enough?

skosito:
@kiwiidb @bumi i implemented some fixes, please check, it seems that tests will fail because there are no funds on lnd3, when that is fixed they should work fine
left comments on constraint changes for negative balance

kiwiidb (changes requested):
Two minor changes requested.
@bumi could comment on the SQL migrations.
Other than that, we should almost be where we want to be. Good work!

db/migrations/20220120000700_add_constraints.up.go
```diff
@@ -34,6 +34,7 @@ func init() {
 				DECLARE
```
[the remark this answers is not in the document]
skosito: @kiwiidb this is `go` migration, bun has support for both `go` and `sql` migrations, i believe one of the reasons for this one is that some code is needed, for example:
```
if db.Dialect().Name().String() != "pg" {
			fmt.Printf("\033[1;31m%s\033[0m", "You are not using PostgreSQL. DB level checks can not be enabled!\n")
			return nil
		}
```
Of course, @bumi will know in more details, but I think this is the only reason for not moving this one to plain `sql`.

lib/service/invoices.go
```diff
@@ -232,7 +234,31 @@ func (svc *LndhubService) HandleSuccessfulPayment(ctx context.Context, invoice *
...
+		Amount:          int64(invoice.Fee),
+		ParentID:        parentEntry.ID,
+	}
+	_, err = svc.DB.NewInsert().Model(&entry).Exec(ctx)
+	if err != nil {
+		sentry.CaptureException(err)
+		svc.Logger.Errorf("Could not insert fee transaction entry user_id:%v invoice_id:%v", invoice.UserID, invoice.ID)
+		return err
+	}
+
```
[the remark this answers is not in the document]
skosito: @kiwiidb done

kiwiidb (approved):

lib/service/invoices.go
```diff
@@ -189,8 +190,10 @@ func (svc *LndhubService) PayInvoice(ctx context.Context, invoice *models.Invoic
 	paymentResponse.TransactionEntry = &entry
 
 	// The payment was successful.
+	// These changes to the invoice are persisted in the `HandleSuccessfulPayment` function
 	invoice.Preimage = paymentResponse.PaymentPreimageStr
-	err = svc.HandleSuccessfulPayment(context.Background(), invoice)
+	invoice.Fee = paymentResponse.PaymentRoute.TotalFees
```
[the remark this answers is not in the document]
kiwiidb: I think it's best for later, I would do a new release as soon as the outstanding PR's are merged.

PR #119, opened 2022-03-03
add build workflow

kiwiidb:
I've only added the binary for linux (default, because it's built using ubuntu) because this should be run on the server, which will in practice always be linux.


kiwiidb:
Sorry, checked out from wrong branch.

PR #120, opened 2022-03-03
add build workflow

kiwiidb:
In practice, lndhub will always run on a linux server, so there is no need for any other binaries.

### DIFF
--- a/.github/workflows/build_binaries.yaml
+++ b/.github/workflows/build_binaries.yaml
@@ -1,8 +1,8 @@
 name: Build and publish binaries
 on:
-  push:
-    tags:
-      - '*'
+  release:
+    types:
+      - created
 jobs:
   build:
     name: Build and publish binaries
@@ -12,6 +12,10 @@ jobs:
         platform: [ ubuntu-latest ]
     runs-on: ${{ matrix.platform }}
     steps:
+    - name: Install Go
+      uses: actions/setup-go@v1
+      with:
+        go-version: ${{ matrix.go-version }}
     - name: Check out code
       uses: actions/checkout@v2
     - name: Build Binary for linux
--- a/.github/workflows/build_binaries.yaml
+++ b/.github/workflows/build_binaries.yaml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ${{ matrix.platform }}
     steps:
     - name: Install Go
-      uses: actions/setup-go@v1
+      uses: actions/setup-go@v2
       with:
         go-version: ${{ matrix.go-version }}
     - name: Check out code
--- a/.github/workflows/build_binaries.yaml
+++ b/.github/workflows/build_binaries.yaml
@@ -18,9 +18,11 @@ jobs:
         go-version: ${{ matrix.go-version }}
     - name: Check out code
       uses: actions/checkout@v2
-    - name: Build Binary for linux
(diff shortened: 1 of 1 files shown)


.github/workflows/build_binaries.yaml
```diff
@@ -0,0 +1,42 @@
...
+      uses: actions/upload-release-asset@v1.0.1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: ${{ github.event.release.upload_url }}
+        asset_path: ./lndhub-${{github.event.release.tag_name}}-linux-x86_64
+        asset_name: lndhub-${{github.event.release.tag_name}}-linux-x86_64
+        asset_content_type: binary/octet-stream
+    - name: Upload armv7 binary to release assets
+      uses: actions/upload-release-asset@v1.0.1
```
bumi, line 35: this seems deprecated: https://github.com/actions/upload-release-asset
they refer to this one as maintained: https://github.com/softprops/action-gh-release

.github/workflows/build_binaries.yaml
```diff
@@ -0,0 +1,42 @@
...
+  build:
+    name: Build and publish binaries
+    strategy:
+      matrix:
+        go-version: [ 1.16.x ]
+        platform: [ ubuntu-latest ]
+    runs-on: ${{ matrix.platform }}
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v1
```
bumi, line 16: there is a v2 or v3 now: https://github.com/marketplace/actions/setup-go-environment

.github/workflows/build_binaries.yaml
```diff
@@ -0,0 +1,42 @@
...
+on:
+  release:
+    types:
+      - created
+jobs:
+  build:
+    name: Build and publish binaries
+    strategy:
+      matrix:
+        go-version: [ 1.16.x ]
```
bumi, line 11: why not `1.17.x`?

.github/workflows/build_binaries.yaml
kiwiidb: If it works, I don't see why we couldn't use it.

kiwiidb: The one you linked to is for creating a release with assets, not uploading assets to an existing one.

.github/workflows/build_binaries.yaml
kiwiidb: Because this PR is for building binaries using our existing Go version, the same one as we use in the Docker build.
If we would want to upgrade our Go version, this is possible but it should go in a different PR. But there is no need for it now.

.github/workflows/build_binaries.yaml
kiwiidb: Ok, updated to v2.

.github/workflows/build_binaries.yaml
kiwiidb: btw, Go 1.18 came out since you posted this comment :)

.github/workflows/build_binaries.yaml
bumi: I just normally like to avoid unmaintained code. 
but works! 👍

bumi (approved):
utACK

PR #123, opened 2022-03-04
Feature/cln

kiwiidb:
This draft PR is meant to track C-Lightning support. According to rumours, @cdecker is working on an official gRPC interface for C-Lightning. So we should rewrite this PR once that is released, so we don't have to rely on the Sparko plugin anymore.


urza:
There was some grpc support aded to CLN in 0.11

https://github.com/ElementsProject/lightning/releases/tag/v0.11.0.1

What exactly is needed?

prusnak:
FTR Core Lightning (fka C-Lightning) has gRPC since version `0.11.0` - see https://github.com/ElementsProject/lightning/releases/tag/v0.11.0.1

prusnak:
Here is the official `cln-grpc` plugin documentation: https://lib.rs/crates/cln-grpc and its API reference: https://docs.rs/cln-grpc/latest/cln_grpc/

kiwiidb:
There are 2 things that are needed right now:

- CLN gRPC interface is missing certain methods, for example we use the `decode` method to decode a bolt11 (I outsourced this to the LN node to reduce dependencies in the lndhub.go code).
- I am waiting on an answer on [this question](https://github.com/ElementsProject/lightning/issues/5227), on how to connect remotely using mTLS.

I will create a new PR for CLN gRPC support and close this one later because it is outdated.

prusnak:
> * CLN gRPC interface is missing certain methods, for example we use the `decode` method to decode a bolt11 (I outsourced this to the LN node to reduce dependencies in the lndhub.go code).

I agree with the sentiment here. However, if CLN gRPC is not adding this, you can use for example this: https://github.com/nbd-wtf/ln-decodepay

kiwiidb:
I've updated this PR but stumbled upon  another issue which blocks CLN integration: C-Lightning doesn’t support creating an invoice with a description_hash as input parameter (it needs the description and does the hashing itself).

PR #127, opened 2022-03-08
Keysend implementation

skosito:


controllers/keysend.ctrl.go
```diff
@@ -0,0 +1,98 @@
+package controllers
```
skosito, line 1: @kiwiidb compared to your changes, I renamed this file to have ctrl suffix and followed your todo hints to implement endpoint

integration_tests/keysend_test.go
```diff
@@ -0,0 +1,122 @@
...
+	// check that balance was reduced
+	userId := getUserIdFromToken(suite.aliceToken)
+	aliceBalance, err := suite.service.CurrentUserBalance(context.Background(), userId)
+	if err != nil {
+		fmt.Printf("Error when getting balance %v\n", err.Error())
+	}
+	assert.Equal(suite.T(), int64(aliceFundingSats)-int64(externalSatRequested), aliceBalance)
+}
+
+func (suite *KeySendTestSuite) TestKeysendPaymentNonExistentDestination() {
```
skosito, line 102: i did not add test with memo too long case, could you please clarify why it's needed @kiwiidb ?

lib/service/invoices.go
```diff
@@ -137,6 +127,44 @@ func (svc *LndhubService) SendPaymentSync(ctx context.Context, invoice *models.I
 	return sendPaymentResponse, nil
 }
 
+func createLnRpcSendRequest(invoice *models.Invoice) (*lnrpc.SendRequest, error) {
```
skosito, line 130: @kiwiidb i unified same code into this method to reduce duplication

lib/service/invoices.go
```diff
@@ -235,7 +263,7 @@ func (svc *LndhubService) HandleSuccessfulPayment(ctx context.Context, invoice *
 	return err
 }
 
-func (svc *LndhubService) AddOutgoingInvoice(ctx context.Context, userID int64, paymentRequest string, decodedInvoice *lnrpc.PayReq) (*models.Invoice, error) {
+func (svc *LndhubService) AddOutgoingInvoice(ctx context.Context, userID int64, paymentRequest string, decodedInvoice *lnrpc.PayReq, keysend bool) (*models.Invoice, error) {
```
skosito, line 266: i think we can maybe add struct that would contain both *lnrpc.PayReq and keysend, just wanted to check if it makes sense from domain perspective?

integration_tests/keysend_test.go
kiwiidb: I think it's fine to leave it out then.

lib/service/invoices.go
kiwiidb: I think it makes sense, yes. Also in the future we might have bolt12 or other extensions of the PayReq. So you can implement it.

main.go
```diff
@@ -134,6 +134,7 @@ func main() {
 	secured.GET("/checkpayment/:payment_hash", controllers.NewCheckPaymentController(svc).CheckPayment)
 	secured.GET("/balance", controllers.NewBalanceController(svc).Balance)
 	secured.GET("/getinfo", controllers.NewGetInfoController(svc).GetInfo)
+	secured.GET("/keysend", controllers.NewKeySendController(svc).KeySend)
```
kiwiidb, line 137: Should be POST.

kiwiidb (changes requested):
- Change to `POST`
- Integrate keysend boolean and lnrpc.PayReq into single struct like you suggested.

If these things are changed, then this would LGTM.

skosito:
> * Change to `POST`
> * Integrate keysend boolean and lnrpc.PayReq into single struct like you suggested.
> 
> If these things are changed, then this would LGTM.

@kiwiidb great catch with GET, not sure how I missed it.
I added a struct, not sure if there is a better name for it, please check.

kiwiidb:
Better to name it "LNPayReq", we don't want extra references to LND because it could hold extra fields that are not related to LND (bolt12).

kiwiidb (approved):

skosito:
> Better to name it "LNPayReq", we don't want extra references to LND because it could hold extra fields that are not related to LND (bolt12).

@kiwiidb sure, renamed it

PR #131, opened 2022-03-10
Rate limiter and cache middleware

skosito:


main.go
```diff
@@ -170,3 +175,33 @@ func main() {
 		e.Logger.Fatal(err)
 	}
 }
+
+func createRateLimitMiddleware(seconds int, burst int) echo.MiddlewareFunc {
+	config := middleware.RateLimiterMemoryStoreConfig{
+		Rate:  rate.Every(time.Duration(seconds) * time.Second),
+		Burst: burst,
+	}
+	return middleware.RateLimiter(middleware.NewRateLimiterMemoryStoreWithConfig(config))
```
skosito, line 184: if i just pass rate.Every(10 * time.Second) that is same as 0.1 and it is treated as 0
only way to do this it seems is to set rate to 0.1, and burst to 1, and it is acting as 1 every 10 seconds

tested it using postman with /create endpoint

main.go
```diff
@@ -143,7 +148,7 @@ func main() {
 
 	//Index page endpoints, no Authorization required
 	homeController := controllers.NewHomeController(svc, indexHtml)
-	e.GET("/", homeController.Home)
+	e.GET("/", homeController.Home, createCacheClient().Middleware())
```
skosito, line 151: tested by checking that only first time home endpoint is hit and then it is cached, please check parameters, i took default

main.go
```diff
@@ -121,14 +124,16 @@ func main() {
...
 
+	onePer10sMiddleware := createRateLimitMiddleware(10, 1)
 	// Public endpoints for account creation and authentication
 	e.POST("/auth", controllers.NewAuthController(svc).Auth)
-	e.POST("/create", controllers.NewCreateUserController(svc).CreateUser)
+	e.POST("/create", controllers.NewCreateUserController(svc).CreateUser, onePer10sMiddleware)
 
 	// Secured endpoints which require a Authorization token (JWT)
-	secured := e.Group("", tokens.Middleware(c.JWTSecret))
+	secured := e.Group("", tokens.Middleware(c.JWTSecret), middleware.RateLimiter(middleware.NewRateLimiterMemoryStore(1)))
```
skosito, line 133: just put one each second for all secured, it can easily be changed with your suggestion

go.sum
```diff
@@ -40,13 +40,17 @@ github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
```
skosito, line 40: after running go mod tidy these indirect dependencies were added

main.go
```diff
@@ -121,14 +124,16 @@ func main() {
 		IdentityPubkey: getInfo.IdentityPubkey,
 	}
 
+	onePer10sMiddleware := createRateLimitMiddleware(10, 1)
 	// Public endpoints for account creation and authentication
 	e.POST("/auth", controllers.NewAuthController(svc).Auth)
```
skosito, line 129: something for auth? maybe doesn't make sense

kiwiidb (changes requested):
Great work, I would just change a couple of things:

- Make the timoutSeconds configurable for rate limits, both the strict limit and the less strict limit. The caching is fine as it is, these params don't need to be in the config.
- Put the `/keysend` endpoint also on the strict rate limit.
- Put the `/getinfo` endpoint under the caching middleware.
- Make the default timeout for the less-strict rate limit more relaxed: 10 requests / second should be fine for calls that don't hit the Lightning node.

skosito:
> * 
> * Put the

@kiwiidb thanks for review, fixed the comments, please check again

kiwiidb (approved):

PR #133, opened 2022-03-18
Feature/keysend-custom-records

kiwiidb:
Allow custom TLV records to be passed on.

The memo field is now internal. If you want to send a Whatsat message, you should specify the TLV record explicitly in th request:

```
http POST http://$host/keysend destination=025c1d5d1b4c983cc6350fc2d756fbb59b4dc365e45e87f8e3afe07e24013e8220 amount:=101 customRecords:='{"34349334":"test_whatsatee"}' Authorization:"Bearer $token"
```

### DIFF
--- a/lib/service/invoices.go
+++ b/lib/service/invoices.go
@@ -157,13 +157,14 @@ func createLnRpcSendRequest(invoice *models.Invoice) (*lnrpc.SendRequest, error)
 	if err != nil {
 		return nil, err
 	}
+	invoice.DestinationCustomRecords[KEYSEND_CUSTOM_RECORD] = preImage
 	return &lnrpc.SendRequest{
 		Dest:              destBytes,
 		Amt:               invoice.Amount,
 		PaymentHash:       pHash.Sum(nil),
 		FeeLimit:          &feeLimit,
 		DestFeatures:      []lnrpc.FeatureBit{lnrpc.FeatureBit_TLV_ONION_REQ},
-		DestCustomRecords: map[uint64][]byte{KEYSEND_CUSTOM_RECORD: preImage, TLV_WHATSAT_MESSAGE: []byte(invoice.Memo)},
+		DestCustomRecords: invoice.DestinationCustomRecords,
 	}, nil
 }
 
--- a/controllers/keysend.ctrl.go
+++ b/controllers/keysend.ctrl.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"fmt"
 	"net/http"
+	"strconv"
 
 	"github.com/getAlby/lndhub.go/lib"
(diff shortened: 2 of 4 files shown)


lib/service/invoices.go
```diff
@@ -157,13 +157,14 @@ func createLnRpcSendRequest(invoice *models.Invoice) (*lnrpc.SendRequest, error)
 	if err != nil {
 		return nil, err
 	}
+	invoice.DestinationCustomRecords[KEYSEND_CUSTOM_RECORD] = preImage
```
bumi, line 160: is this now always set? not only on keysend payements?

lib/service/invoices.go
```diff
@@ -157,13 +157,14 @@ func createLnRpcSendRequest(invoice *models.Invoice) (*lnrpc.SendRequest, error)
...
 	return &lnrpc.SendRequest{
 		Dest:              destBytes,
 		Amt:               invoice.Amount,
 		PaymentHash:       pHash.Sum(nil),
 		FeeLimit:          &feeLimit,
 		DestFeatures:      []lnrpc.FeatureBit{lnrpc.FeatureBit_TLV_ONION_REQ},
-		DestCustomRecords: map[uint64][]byte{KEYSEND_CUSTOM_RECORD: preImage, TLV_WHATSAT_MESSAGE: []byte(invoice.Memo)},
```
bumi, line 166: memo is now no longer added as default TLV_WHATSAT_MESSAGE?

lib/service/invoices.go
kiwiidb: On line 143, if the invoice is not keysend, we have an early return.

lib/service/invoices.go
kiwiidb: No, I've decided to leave this up to the client.
I've left the memo field to keep backwards compatibility, but if it is not added explicitly to the custom records by the client, it's only purpose is to be stored in the database as an internal memo field.

lib/service/invoices.go
bumi: ah, I missed that. ok, thanks.

in such code branches I prefer using if/else to make it clear

kiwiidb: I prefer the early return, as described in the [Zen of Go](https://the-zen-of-go.netlify.app/).

lib/service/invoices.go
bumi: ok. 👍

lib/service/invoices.go
bumi: :+1:
(just might be more easy to miss, but 🐐 )

bumi (approved):
some specs would be nice maybe

PR #134, opened 2022-03-18
Unauthorized invoice endpoint

skosito:
Used the same logic as addInvoice, just getting user from db first instead of getting it from token.

### DIFF
--- a/controllers/addinvoice.ctrl.go
+++ b/controllers/addinvoice.ctrl.go
@@ -33,6 +33,10 @@ type AddInvoiceResponseBody struct {
 // AddInvoice : Add invoice Controller
 func (controller *AddInvoiceController) AddInvoice(c echo.Context) error {
 	userID := c.Get("UserID").(int64)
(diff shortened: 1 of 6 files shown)


bumi:
@godSaysHODL can you have a look? is this how you imagine it?

godSaysHODL:
Yes, it looks great from my very untrained eyes 👁️👁️🙌

bumi (commented):
we should add a spec

kiwiidb:
What do you mean by "spec" ?

bumi:
sorry, I mean "tests".

(in ruby we often use specs)

skosito:
> sorry, I mean "tests".
> 
> (in ruby we often use specs)

sure, good point, i will add tests, and then we can continue review

kiwiidb:
Looks good, some basic tests and then this can be merged

skosito:
@bumi @kiwiidb I added simple tests, please check again.

kiwiidb (approved):

PR #136, opened 2022-03-23
Dynamic fee limits

skosito:


lib/service/invoices.go
```diff
@@ -168,6 +161,19 @@ func createLnRpcSendRequest(invoice *models.Invoice) (*lnrpc.SendRequest, error)
 	}, nil
 }
 
+func calcFeeLimit(invoice *models.Invoice) lnrpc.FeeLimit {
+	limit := int64(10)
```
skosito, line 165: put these values here based on issue description
is there a way to test this somehow? currently test nodes are not responding, so I can not run existing ones, when that is stable i will verify and potentially add more

bumi: Are the test nodes working again? @kiwiidb

kiwiidb: Tests have passed now.

bumi: I think a test for the calcFeeLimit function would be good.

skosito: @bumi i added couple unit tests for this function, please check

kiwiidb: OK

kiwiidb (approved):

PR #137, opened 2022-03-23
Fix/pay-response

kiwiidb:
Addresses the bug where `payment_route` in the pay responses was empty. We weren't following the Bluewallet spec here.


bumi (approved):
just confirmed also. this is the correct lndhub response.

PR #138, opened 2022-03-23
Started API docs

bumi:


kiwiidb:
Wouldn't it be better to use something like https://github.com/swaggo/echo-swagger?

kiwiidb:
Superseded by #161

PR #146, opened 2022-04-05
Do not send bad auth errors to sentry

skosito:


lib/responses/errors.go
```diff
@@ -58,3 +59,32 @@ func HTTPErrorHandler(err error, c echo.Context) {
 	}
 	// TODO: use an error matching the error code
 }
+
+// this is a simple way to try to convert err.Message interface
+// to ErrorResponse without external packages
+func errToErrorResponse(err error) *ErrorResponse {
```
skosito, line 65: since this is map, i think converting to json and back to struct is the only built in way in golang to do this
there is package for decoding struct, but i did not want to add it just for this

bumi:
hmm. actually I was wondering about the HTTP 400 code... 
we probably should return a 401 there. 
https://github.com/getAlby/lndhub.go/blob/main/lib/tokens/jwt.go#L28

Do you know if the sentry middleware somehow defines which error codes it sends? they do that in ruby - e.g. by default 401 is ignored by the sentry middleware

skosito:
> hmm. actually I was wondering about the HTTP 400 code... we probably should return a 401 there. https://github.com/getAlby/lndhub.go/blob/main/lib/tokens/jwt.go#L28
> 
> Do you know if the sentry middleware somehow defines which error codes it sends? they do that in ruby - e.g. by default 401 is ignored by the sentry middleware

Oh ok, I will check, thanks!

main.go
```diff
@@ -93,7 +93,8 @@ func main() {
 	// Setup exception tracking with Sentry if configured
 	if c.SentryDSN != "" {
 		if err = sentry.Init(sentry.ClientOptions{
-			Dsn: c.SentryDSN,
+			Dsn:          c.SentryDSN,
+			IgnoreErrors: []string{"401"},
```
skosito, line 97: this only accept error of regex, but setting this up like this ignores 401 errors, tested out on free sentry account i made

skosito:
> hmm. actually I was wondering about the HTTP 400 code... we probably should return a 401 there. https://github.com/getAlby/lndhub.go/blob/main/lib/tokens/jwt.go#L28
> 
> Do you know if the sentry middleware somehow defines which error codes it sends? they do that in ruby - e.g. by default 401 is ignored by the sentry middleware

I changed it with adding IgnoreErrors field, please check again

bumi (approved):
bäm! this looks good.

bumi:
mergeable once tests are updated

kiwiidb:
Looks good, I'm merging.

PR #149, opened 2022-04-08
always lowercase payreq

kiwiidb:
When testing out Alby and Bluewallet on regtest, I noticed that Bluewallet puts in the invoice uppercase when calling `payinvoice`. This caused internal Alby payments to fail, as Alby could not find the invoice in the db.

### DIFF
--- a/controllers/payinvoice.ctrl.go
+++ b/controllers/payinvoice.ctrl.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/getAlby/lndhub.go/lib"
 	"github.com/getAlby/lndhub.go/lib/responses"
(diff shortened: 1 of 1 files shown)


bumi:
Can you explain that? why is it relevant that we pass in the invoice in lowercase to the DecodePayReq call? 
if we always need it lowercase shouldn't we do it then here: https://github.com/getAlby/lndhub.go/blob/8ed306fe61370beb484530a646bd6a7b90a93145/lib/service/invoices.go#L392 ?

bumi (approved):

kiwiidb:
This variable is not only passed into the `DecodePayReq` call, but also below, in the `AddOutgoingInvoice` call. This causes a mismatch between lowercase and an uppercase payment request strings in the database.

bumi:
ok, but then we should do it there. Making sure that we write lowercase in the DB and also use lowecase when we try too look it up. 
Maybe `AddOutgoingInvoice` will be used somewhere else then we have the same problem again.

PR #150, opened 2022-04-11
deploy canary on new release

kiwiidb:
Like this a new tag (release or tag pushed) will be auto-deployed in production on a seperate, canary deployment `lndhub-canary.mainnet.getalby.com`. This will allow us to do a test in production without pushing the release out to all users.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -46,7 +46,7 @@ jobs:
       uses: fjogeleit/yaml-update-action@v0.7.0
       with:
         valueFile: 'alby-mainnet-deployment/values.yaml'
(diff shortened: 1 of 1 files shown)


bumi:
what is this change doing? how does this affect the production deployment?

kiwiidb:
I thought it would be good to not automatically deploy a new release to all users, but to test it out ourselves in production first, and then roll out to all users manually. But since I have removed the canary deployment because of the double-transaction issue, this PR is not useful at the moment. We could still have a canary deployment in production with a seperate database though.

kiwiidb:
As the canary release has been deleted, this is no longer relevant.

PR #152, opened 2022-04-13
Replace math rand with crypto rand

skosito:
I just did plain replace and extracted to common method. 

Just a though - even though math rand is not secure, maybe for these 2 use cases where it just finds random value between 2 numbers using predefined string of alphanumberic characters there are no security implications and using math rand is ok? Using crypto rand however is pretty much the same, only difference is that it can return an error, so i refactored a bit around that.

Base is websocket branch because it is using rand code in Subscribe, so just to avoid conflicts after merge.


bumi:
@kiwiidb is this OK?

kiwiidb:
Fixed conflicts, let's wait for tests and then it's ok.
Right now, `math/rand` is seeded with the Unix nano time at the moment LNDhub is spinned up. Anyone who has knowledge of that number could compromise everyone's account. That's why it's better to have the `crypto/rand`.

PR #156, opened 2022-04-21
Publish invoice for internal payments as well

kiwiidb:


kiwiidb:
> how about keysend?
Receiving Keysend payments is currently not supported in LNDhub in any way. FYI what we have implemented here is a pubsub system only for incoming payments, not outgoing.

bumi:
ah sure sorry... 😵‍💫 :)

kiwiidb:
I'm trying to get a succesful test run in. But we should look into booting regtest networks inside the Github Action pipeline, maybe that helps to get more consistent test results.

bumi (approved):

PR #157, opened 2022-04-21
Add logo and update favicon

### LINKED ISSUE #88: Create a logo :)

stefanwuest:
Closes #88 
Adding the logo to the repo and updating the favicon based on the logo.


kiwiidb:
Should we also add this to the top of the `index.html` page? So the branding is more neutral and not referring to Alby anymore?

bumi:
yes, though ideally it would be configurable there maybe? - but that's too complex and not worth it

PR #159, opened 2022-04-22
background routine: cancel when program ends

### LINKED ISSUE #158: Graceful shutdown of invoice processing routine
Currently, we do not cancel the context of the background routine. I've already added support for canceling the context due to the need for this in the integration tests. But we should do it in the main routine as well.

kiwiidb:
Fixes #158

### DIFF
--- a/main.go
+++ b/main.go
@@ -166,7 +166,8 @@ func main() {
 	e.GET("/invoices/stream", controllers.NewInvoiceStreamController(svc).StreamInvoices)
 
 	// Subscribe to LND invoice updates in the background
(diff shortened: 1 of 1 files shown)


bumi (approved):
(ACK with not enought Go knowledge :)

prusnak:
Should this be merged?

kiwiidb:
I was testing this and came to the conclusion that the shutdown was still not handled gracefully, even with this change. So I need to do some more testing.

kiwiidb:
Will be superseded by #291

PR #162, opened 2022-04-27
Feature/webhooks

kiwiidb:
Fixes #162 

- Change pubsub topic type from int to string
- Publish both incoming and outgoing invoices
- Subscribe to both incoming and outgoing invoices in a new goroutine
- Send  out the invoices as a POST request
- Add test


main.go
```diff
@@ -168,6 +168,13 @@ func main() {
 	// Subscribe to LND invoice updates in the background
 	go svc.InvoiceUpdateSubscription(context.Background())
 
+	//Start webhook subscription
+	if svc.Config.WebhookUrl != "" {
+		webhookCtx, cancelWebhook := context.WithCancel(context.Background())
+		go svc.StartWebhookSubscribtion(webhookCtx)
+		defer cancelWebhook()
```
bumi, line 175: Do I understand this Go correct: 
Sending the webhook basically happens asynchronously in its own webhook context? 
This means that the HTTP response finishes/returns and the webhook still executes?

lib/service/webhook.go
```diff
@@ -0,0 +1,53 @@
...
+		return
+	}
+
+	resp, err := http.Post(svc.Config.WebhookUrl, "application/json", payload)
+	if err != nil {
+		svc.Logger.Error(err)
+		return
+	}
+	if resp.StatusCode != http.StatusOK {
+		msg, err := ioutil.ReadAll(resp.Body)
```
bumi, line 47: wondering if we actually need the body. It's great for debugging but should it be done in production?

main.go
```diff
@@ -168,6 +168,13 @@ func main() {
 	// Subscribe to LND invoice updates in the background
 	go svc.InvoiceUpdateSubscription(context.Background())
 
+	//Start webhook subscription
+	if svc.Config.WebhookUrl != "" {
+		webhookCtx, cancelWebhook := context.WithCancel(context.Background())
+		go svc.StartWebhookSubscribtion(webhookCtx)
```
bumi, line 174: from the implementation of `StartWebhookSubscription` this could be called multiple times? 
if so could the URL also passed in as argument?

bumi (approved):
nice.

bumi (commented):
readme update is missing

main.go
kiwiidb: Yes, this is totally seperate from the main server. LNDhub has it's own Pubsub system where `models.Invoice` objects are being passed around that represent ingoing and outgoing invoices.

lib/service/webhook.go
```diff
@@ -0,0 +1,53 @@
...
+		svc.Logger.Error(err)
+		return
+	}
+
+	resp, err := http.Post(svc.Config.WebhookUrl, "application/json", payload)
+	if err != nil {
+		svc.Logger.Error(err)
+		return
+	}
+	if resp.StatusCode != http.StatusOK {
```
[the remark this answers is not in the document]
kiwiidb: There's not a really elegant way of doing this (you would nee to do a literal 200 =< resp.StatusCode < 300) We could mandate that the callback needs to return exactly `200`. Seems not that demanding since the receiving server also has to parse our `Invoice` object. But I could add a `201` as a valid response code, seems like the rest of the 200 range is more exotic.

main.go
kiwiidb: When you make a `Pay` call, the publishing part of the PubSub system is executed synchronously, but all possible subscribers are executed asynchronously.

lib/service/webhook.go
kiwiidb: I would still log it in case we get an error though, doesn't really add any overhead.

bumi: ok, depends on what the server responds :) at least the logs will get full.

main.go
kiwiidb: Right now this is only called a single time, here in `main.go`. But we could call it once per webhook in the future. This could be okay in case there is a limited amount of webhooks (probably the case). The other option is to have a single subscription/goroutine for the webhook system, and call the `postToWebhook` multiple times.

README.md
```diff
@@ -77,6 +78,37 @@ LndHub.go supports PostgreSQL and SQLite as database backend. But SQLite does no
...
+  "state": "settled",
+  "error_message": "",
+  "add_index": 0,
+  "CreatedAt": "2022-04-27T13:50:43.938597+02:00",
+  "ExpiresAt": "2022-04-27T14:49:37+02:00",
+  "updated_at": "2022-04-27T13:50:44.313549+02:00",
+  "settled_at": "2022-04-27T13:50:44.313539+02:00"
+}
+```
+
```
kiwiidb, line 111: @bumi what do you think about this payload?

bumi: hmm. needs to be unified. `updated_at` vs. `CreatedAt` 

I'd remove `User` - we have `user_id` which is enough.

what will we have in `DestinationCustomRecords`? 

is `add_index` and `internal` needed?

kiwiidb: I have unified the json fields with snake case everywhere. User, add_index and internal are removed, and an example of `custom_records` has been added to the README.

main.go
bumi: 👍

bumi (approved):

PR #167, opened 2022-05-02
Fix: 0 amount incoming invoice

### LINKED ISSUE #165: Add support for 0-amount incoming invoices
Currently 0-amount invoices are not supported and are not properly settled. 

When the transaction is settled we currently use the invoice.Amount (which is 0): 
https://github.com/getAlby/lndhub.go/blob/feature/webhooks/lib/service/invoicesubscription.go#L85

we should use the amount from the `rawInvoice` there but log if the amount is not the same as the invoice.Amount. 
Maybe we also only do this if the invoice Amount is zero?

kiwiidb:
Fixes #165 

This credits the user's account the full amount that was paid for all incoming invoices. When there is a mismatch, a line is logged.
I pushed an integration test that reproduces the bug in the first commit.

### DIFF
--- a/lib/service/invoicesubscription.go
+++ b/lib/service/invoicesubscription.go
@@ -81,7 +81,11 @@ func (svc *LndhubService) ProcessInvoiceUpdate(ctx context.Context, rawInvoice *
 			InvoiceID:       invoice.ID,
 			CreditAccountID: creditAccount.ID,
 			DebitAccountID:  debitAccount.ID,
-			Amount:          invoice.Amount,
+			Amount:          rawInvoice.AmtPaidSat,
+		}
+
+		if rawInvoice.AmtPaidSat != invoice.Amount {
+			svc.Logger.Infof("Incoming invoice amount mismatch. Invoice id:%v, amt: %d, amt_paid: %d.", invoice.ID, invoice.Amount, rawInvoice.AmtPaidSat)
 		}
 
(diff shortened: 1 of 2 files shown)


lib/service/invoicesubscription.go
```diff
@@ -81,7 +81,11 @@ func (svc *LndhubService) ProcessInvoiceUpdate(ctx context.Context, rawInvoice *
 			InvoiceID:       invoice.ID,
 			CreditAccountID: creditAccount.ID,
 			DebitAccountID:  debitAccount.ID,
-			Amount:          invoice.Amount,
+			Amount:          rawInvoice.AmtPaidSat,
+		}
+
+		if rawInvoice.AmtPaidSat != invoice.Amount {
+			svc.Logger.Infof("Incoming invoice amount mismatch. Invoice id:%v, amt: %d, amt_paid: %d.", invoice.ID, invoice.Amount, rawInvoice.AmtPaidSat)
```
bumi, line 88:
```suggestion
			svc.Logger.Infof("Incoming invoice amount mismatch. user_id:%v invoice_id:%v, amt:%d, amt_paid:%d.", invoice.UserID, invoice.ID, invoice.Amount, rawInvoice.AmtPaidSat)
```

bumi (approved):
I guess we can trust `rawInvoice` ? :)

PR #177, opened 2022-05-24
Fix: start invoice subscription from last NOT expired invoice

bumi:
Currently we checked all payments which means we start the subscription from an index for an invoice that can no longer be paid because it is expired.
And because it is expired we do not need to listen for updates for it anymore

### DIFF
--- a/integration_tests/subscription_start_test.go
+++ b/integration_tests/subscription_start_test.go
@@ -1,0 +1,139 @@
+package integration_tests
+
+import (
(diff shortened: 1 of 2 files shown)


kiwiidb:
I added a test that fails if you remove the fix.

kiwiidb (approved):

bumi (commented):
from what I understand, looks good!

PR #187, opened 2022-06-08
Fix/keysend-payments

kiwiidb:
When doing keysend payments, outgoing payments would not have an `r_hash`. This change fixes that.

Also removes the default memo for keysend payments and makes it blank.


bumi (approved):
utACK

not sure if tests are useful?

PR #189, opened 2022-06-10
Fix/amt-validations

kiwiidb:
This fixes some bugs where users would be able to add invoices for negative amounts to the database, or to try and keysend negative amounts.

### DIFF
--- a/controllers/addinvoice.ctrl.go
+++ b/controllers/addinvoice.ctrl.go
@@ -61,7 +61,7 @@ func AddInvoice(c echo.Context, svc *service.LndhubService, userID int64) error 
 	}
 
 	amount, err := svc.ParseInt(body.Amount)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


controllers/addinvoice.ctrl.go
```diff
@@ -61,7 +61,7 @@ func AddInvoice(c echo.Context, svc *service.LndhubService, userID int64) error
 	}
 
 	amount, err := svc.ParseInt(body.Amount)
-	if err != nil {
```
bumi, line 64: we need to check this here because we accept strings as request body?

kiwiidb: The validate tag won't work because it's a string right? So we have to check seperately

bumi: true, thanks.

PR #193, opened 2022-06-12
Implement fee-reserve

prusnak:
My shot at fixing/implementing https://github.com/getAlby/lndhub.go/issues/183

It's rather naive and I am not sure if having the check at one place only is enough.

Someone with more Go expertise than me should carefully review.

And yeah, this PR does not contain any tests ... :-/

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,10 +16,6 @@ Live deployment at [ln.getalby.com](https://ln.getalby.com).
 
 ### Status: alpha
 
-## Known Issues
-
-* Fee reserves are not checked prior to making the payment. This can cause a user's balance to go below 0.
-
 ## Configuration
 
 All required configuration is done with environment variables and a `.env` file can be used.
@@ -50,6 +46,7 @@ vim .env # edit your config
 + `ENABLE_PROMETHEUS`: (default: false) Enable Prometheus metrics to be exposed
 + `PROMETHEUS_PORT`: (default: 9092) Prometheus port (path: `/metrics`)
 + `WEBHOOK_URL`: Optional. Callback URL for incoming and outgoing payment events, see below.
++ `FEE_RESERVE`: (default: false) Keep fee reserve for each user
 
 ## Developing
 
--- a/controllers/keysend.ctrl.go
+++ b/controllers/keysend.ctrl.go
@@ -85,7 +85,11 @@ func (controller *KeySendController) KeySend(c echo.Context) error {
 		return err
 	}
 
-	if currentBalance < invoice.Amount {
+	minimumBalance := invoice.Amount
+	if controller.svc.Config.FeeReserve {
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -50,6 +50,7 @@ vim .env # edit your config
 + `ENABLE_PROMETHEUS`: (default: false) Enable Prometheus metrics to be exposed
 + `PROMETHEUS_PORT`: (default: 9092) Prometheus port (path: `/metrics`)
 + `WEBHOOK_URL`: Optional. Callback URL for incoming and outgoing payment events, see below.
++ `FEE_RESERVE`: (default: 0) Keep fee reserve for each user (in satoshi)
```
prusnak, line 53: Are balances kept in satoshi or milli-satoshi?

kiwiidb (commented):
Thanks for your PR! The fee-reserve topic is not easy and it's something that we have debated internally for quite a bit already. Some remarks about this PR:

- users with a balance below `Config.FeeReserve` would be unable to send any payments. 
-  it would still be possible for users to go below 0 because the fee of a large enough payment will be larger than any fixed reserve.
- (Unrelated) the balance check actually already happens [here](https://github.com/getAlby/lndhub.go/blob/main/controllers/payinvoice.ctrl.go#L98) and [here](https://github.com/getAlby/lndhub.go/blob/main/controllers/keysend.ctrl.go#L88) (code duplication I know), so it's better to do the check there.

I agree that we can make it configurable for operators if they want a fee reserve or not. But if they want a fee reserve, this should just be a boolean option and the fee reserve should be exactly the same as the fee limit, which is calculated [here](https://github.com/getAlby/lndhub.go/blob/main/lib/service/invoices.go#L179). Adding more numerical limits makes things too complex imo.

prusnak:
> The fee-reserve topic is not easy and it's something that we have debated internally for quite a bit already.

Yeah, I totally get that this is a complex topic.

I addressed your comments and force-pushed 361ae94

kiwiidb (changes requested):
There are some static check failures because of the now-undefined `calcFeeLimit(invoice)` function in `invoices_test.go`, and .
Other than that LGTM, since this is opt-in, I could implement tests in a seperate PR.

kiwiidb (approved):
I approve. @bumi gonna leave this open for a couple of days so you can have a look.

bumi (approved):
yeah, nice! I think that's a great and simple addition for this topic! thanks!

a test would be nice? sending 100% of a user's balance with fee-reserve enabled should fail.

kiwiidb:
Tests are currently failing because they run against our regtest LND nodes, which seem to have some issues the last couple of days. I'm looking into replacing the LND connection with an in-memory mock for the integration tests.

prusnak:
Rebased and force-pushed

PR #197, opened 2022-06-15
Chore/mock lnd

### LINKED ISSUE #195: Mock all LND clients
Currently our LND simnetwork is not stable enough to be used as a back-end for all these integration tests.

kiwiidb:
Fixes #195 

This PR will remove all dependencies on real regtest LND nodes in the Alby regtest cluster.
It replaces them with an in-memory mock of an LND node, which can be found in `integration_tests/lnd_mock.go`.
This mock can serialize and deserialize `bolt11` invoices. For this purpose we have re-introduced the dependency on `btcd`.


kiwiidb:
LND depends heavily on libraries from `btcd`, eg. for signature generation. Bolt11 encoding/decoding includes signature generation/verification, so that's why we need it. By dependency I mean a library dependency, not a dependency on a `btcd` daemon, to be totally clear.

PR #198, opened 2022-06-16
Make it possible to disable account creation

prusnak:
This covers the usecase when LndHub is used to serve closed communities
which do not want to accept new members anymore (friends, families, etc).

The PR introduces a new envconfig option CREATE_ACCOUNTS which is true
by default but can be set to false if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ vim .env # edit your config
 + `PROMETHEUS_PORT`: (default: 9092) Prometheus port (path: `/metrics`)
 + `WEBHOOK_URL`: Optional. Callback URL for incoming and outgoing payment events, see below.
 + `FEE_RESERVE`: (default: false) Keep fee reserve for each user
++ `ALLOW_ACCOUNT_CREATION`: (default: true) Enable creation of new accounts
 
 ## Developing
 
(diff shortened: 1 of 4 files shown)


bumi:
great!
I am wondering if we should define some admin API key for that? Instead of a boolean we could define some kind of "secret / api token" that is required when calling the `/create` endpoint. 

This would allow admins to still create accounts but the endpoint is not publicly available.
This is also helpful when you run lndhub as part of an "micro-service" architecture. (e.g. creating accounts for users from within another application)

prusnak:
> Instead of a boolean we could define some kind of "secret / api token" that is required when calling the `/create` endpoint.

Thats an interesting idea, but I still would sleep better if there was an option to disable this feature completely. Maybe by setting the secret to some particular value (e.g. "DISABLED")?

bumi:
OK. I am fine with both. Is that a good pattern?
- we can also add an admin API token to limit the account creation endpoint in a future PR.

If we keep the boolean: what do you think about naming the env variable `ALLOW_ACCOUNT_CREATION`?

How would create accounts if it is disabled? The workflow would be to first have it enabled, create certain accounts and then disable it?

prusnak:
Renamed the variable to `ALLOW_ACCOUNT_CREATION`

I agree that the secret API token can be added in the later PR.

bumi (approved):

kiwiidb:
Looks good. In an enterprise environment you could also solve this at an ingress level. But this would indeed be useful for personal/family use.

PR #201, opened 2022-06-17
fix typo in error message

prusnak:
Tested with BlueWallet and ZeusLN and both show 1%% instead of 1%


kiwiidb:
I also noticed this, and have already fixed the typo as part of #202 .

PR #209, opened 2022-07-10
Fix version in documentation

prusnak:
* There are lot of places where we are stuck at 0.6.1 while the latest version is 0.9.0
* Fix license identifier (GPL -> GPLv3) to make it more obvious which version is being used


bumi:
👍 could the versions be loaded from a constant? 
I fear that we will miss those again :)

prusnak:
Not sure, there is just one place that is actually the code. The rest is a comment, json and yaml.

PR #211, opened 2022-07-11
Fix typo and whitespace

prusnak:


bumi (commented):
oooppps... I am horrible at seeing typos, should have seen this in the review.

bumi (approved):

README.md
```diff
@@ -75,7 +75,7 @@ Alternatively you can also use the [Alby simnetwork](https://github.com/getAlby/
 
 ## Database
 
-LndHub.go requires a PostgreSQL database backend.
+LNDhub.go requires a PostgreSQL database backend.
```
bumi, line 78: LNDhub or LndHub? :) 

hehe, everytime I write lndhub I look somewhere to see how it's written... and probably everytime I look somewhere where it is different :D

prusnak: I used LndHub for upstream project and LNDhub.go for this project, because that's how it is used in most of the places. 

We can use LndHub.go if you want but then we'd need to regenerate lndhub.png from the README.md and I don't know how to do that ;-)

prusnak: And finally, there is a third option ... 

If there is a still plan to support CLN (https://github.com/getAlby/lndhub.go/pull/123) we might as well rename the whole project to something completely new, but that option opens a completely new can of worms :-)

prusnak: Giving it more thought today - I updated the PR to use `LndHub` and `LndHub.go` constistently and simplified the logo (to not contain the text) and the README. I think that's the best way to proceed and also opens the door in case we want to rename the project since the text is no longer part of the logo.

bumi (approved):

kiwiidb (approved):

Issue #113, opened 2022-02-18
Make homepage template configurable


bumi:
Currently we embed the homepage template. The template contains Alby specific information. 
Maybe it makes sense to make the the HTML template configurable. 
This could be done by optionally loading the template from an configurable path.

bumi:
ref: #212

PR #212, opened 2022-07-11
Make it possible to override branding

prusnak:
This PR builds on top of https://github.com/getAlby/lndhub.go/pull/211

Fixes https://github.com/getAlby/lndhub.go/issues/113


lib/service/config.go
```diff
@@ -1,27 +1,56 @@
...
+	FeeReserve            bool          `envconfig:"FEE_RESERVE" default:"false"`
+	AllowAccountCreation  bool          `envconfig:"ALLOW_ACCOUNT_CREATION" default:"true"`
+	MaxReceiveAmount      int64         `envconfig:"MAX_RECEIVE_AMOUNT" default:"0"`
+	MaxSendAmount         int64         `envconfig:"MAX_SEND_AMOUNT" default:"0"`
+	MaxAccountBalance     int64         `envconfig:"MAX_ACCOUNT_BALANCE" default:"0"`
+	BrandingTitle         string        `envconfig:"BRANDING_TITLE" default:"LNDhub.go - Alby Lightning"`
+	BrandingDesc          string        `envconfig:"BRANDING_DESC" default:"Alby server for the Lightning Network"`
+	BrandingUrl           string        `envconfig:"BRANDING_URL" default:"https://ln.getalby.com"`
+	BrandingLogo          string        `envconfig:"BRANDING_LOGO" default:""`
+	BrandingFavicon       string        `envconfig:"BRANDING_FAVICON" default:"../static/img/favicon.png"`
```
bumi, line 36: that actually should be /static/img/favicon.png :) - do we even need a favicon? 

customizing those is actually a bit of a problem because right now everything is embedded in the compiled lndhub executable. 
When those are customized those would need to be served from somewhere else.
To configure those a full URL would need to be used. 

wondering if we could simplify this? 

For example: 
IF some config option to a directory is set. then we serve the static files in that directory and use the index.html in that folder as template. 

just thinking out load. 

but +1 for having this configurable! when people run their own lndhub the Alby branding is confusing.

prusnak: > To configure those a full URL would need to be used.

You are right. I thought that `data:image/png;base64,....` would work just fine, but I run into https://github.com/golang/go/issues/20536 which forbids that for 

> IF some config option to a directory is set. then we serve the static files in that directory and use the index.html in that folder as template.

What I like about my approach is that it still could work in a deployment where `lndhub` binary does not have an access to the filesystem (just sockets). If that is out of scope, we can can use your solution and create an env arguments which overrides the whole`templates/index.html`.

Let me know what you think and I'll update the PR.

prusnak:
Rebased on top of current `main` branch after https://github.com/getAlby/lndhub.go/pull/211 merge

kiwiidb (approved):
Looks good. Maybe prusnak-themed branding as example? ;-)

PR #215, opened 2022-07-21
Enable "unix:" prefix for PostgreSQL connection string

prusnak:
According to Pgdriver documentation a valid connection string can also start with `unix://` when UNIX sockets are used. See https://bun.uptrace.dev/postgres/#pgdriver

This PR updates the checks to allow that.


bumi (approved):
utACK

kiwiidb (approved):

PR #217, opened 2022-07-22
Lnurlp + Auth middleware

juligasa:
I am pulling a bunch of changes that we need to get the lnurl working. Mostly we need lnurlp and expand the /create route functionality
so it accepts further credentials.

- LNURLP: need a new route so clients can call it and get the proper response. In the response [callback param](https://github.com/fiatjaf/lnurl-rfc/blob/luds/06.md#wallet-to-service-interaction-flow) we set /invoice so the LN WALLET will call that route to get the invoice. Hence I needed to modify that route so it does not need credentials (just a valid login/nickname)
- NICKNAME: Since we will support lightning addresses in the future, I added a unique identifier of a user that, unlike login, could be updated. The way the nickname is updated is calling /create with a valid (and already created) login & password, producing a postgres index collision that updates the nickname. Additionally you can search a user by both login and nickname.
- CREATEV2: I needed a new way of creating an account that fits our needs, a way that only verifiable peers can do. If the login provided in the /create route is a libp2p CID (we can check that) then the password must be a signature and user must provide the public key of that signature. That's why I introduced a pubkey in the Authorization header so the client signs a message (typically "log in into service bla bla") with its private key and puts the signature as a password in the /create route. It also includes the public key in the authorization header (for those schemas that pubkey cannot be derived from the signature). So the new signature authorization middleware verifies that the signature is valid when the login is a CID. If not, then the extra validation is skipped and everything is as before. The last step in this new validation is validating that the pubkey yields to the login provided.
- NITS AND IMPROVEMENTS: I have also found various nits and also grow the config variables so it is more flexible to operate the program

Maybe there are a lot of changes, but I just wanted to pull everything so we can decide if there is anything relevant at all🙂. In that case I can try to isolate those changes


lib/service/config.go
```diff
@@ -13,7 +13,9 @@ type Config struct {
 	JWTRefreshTokenExpiry int    `envconfig:"JWT_REFRESH_EXPIRY" default:"604800"` // in seconds, default 7 days
 	JWTAccessTokenExpiry  int    `envconfig:"JWT_ACCESS_EXPIRY" default:"172800"`  // in seconds, default 2 days
 	LNDAddress            string `envconfig:"LND_ADDRESS" required:"true"`
-	LNDMacaroonHex        string `envconfig:"LND_MACAROON_HEX" required:"true"`
+	LNDMacaroonFile       string `envconfig:"LND_MACAROON_FILE"`
+	LNDCertFile           string `envconfig:"LND_CERT_FILE"`
```
prusnak, line 17: I was planning to implement these two as well. It is a perfect candidate to be extracted into a separate PR.

lib/security/password.go
```diff
@@ -1,13 +1,14 @@
...
+	"crypto/sha256"
+	"encoding/hex"
 )
 
 // HashPassword : Hash Password
 func HashPassword(password string) string {
-	bytes, _ := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
-	password = string(bytes)
+	bytes := sha256.Sum256([]byte(password))
+	password = hex.EncodeToString(bytes[:])
```
prusnak, line 11: This weakens the security, since one can use rainbow-tables or simple Google search to decode the password. For example `5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8` is easily to be revealed as `password` via Google search.

At least use HMAC-SHA256 with constant hash, which is harder to be reverted. Something like:

```
secret := "LndHub.go is the best"
h := hmac.New(sha256.New, []byte(secret))
h.Write([]byte(password))
password = hex.EncodeToString(h.Sum(nil))
```

lib/service/config.go
prusnak: Extracted into https://github.com/getAlby/lndhub.go/pull/218

db/migrations/20220712131700_nicknames.up.sql
```diff
@@ -0,0 +1,9 @@
+alter table users ADD COLUMN nickname character varying NOT NULL;
```
juligasa, line 1: When migrating an old database this yields to an error like `"Error migrating database: ERROR: column \"nickname\" of relation \"users\" contains null values (SQLSTATE=23502)"}`

lib/security/password.go
juligasa: Yeah, makes sense

db/migrations/20220712131700_nicknames.up.sql
prusnak: I guess you need to drop `NOT NULL`?

juligasa: Yes, but since `nickname` will be the [lightning address username](https://github.com/andrerfneves/lightning-address/blob/master/README.md) those old users without `nickname` can only be paid by its `login` (since there is now a function  that search users by login OR nickname) which is no big deal I guess ( given that logins are pretty ugly strings :) )

integration_tests/util.go
```diff
@@ -55,7 +55,7 @@ func LndHubTestServiceInit(lndClientMock lnd.LightningClientWrapper) (svc *servi
 		MaxReceiveAmount:      1000000,
 		MaxSendAmount:         100000,
 		LnurlAPIPrefix:        "testnet",
-		LnurlDomain:           "mintter.com",
+		LnurlDomain:           "domain.com",
```
prusnak, line 58: `example.com` is better since it is reserved - see https://en.wikipedia.org/wiki/Example.com - while http://domain.com is a commercial service

juligasa: done!

integration_tests/util.go
```diff
@@ -52,6 +52,10 @@ func LndHubTestServiceInit(lndClientMock lnd.LightningClientWrapper) (svc *servi
 		JWTRefreshTokenExpiry: 3600,
 		LNDAddress:            mockLNDAddress,
 		LNDMacaroonHex:        mockLNDMacaroonHex,
+		MaxReceiveAmount:      1000000,
+		MaxSendAmount:         100000,
+		LnurlAPIPrefix:        "testnet",
+		LnurlDomain:           "example.com",
```
prusnak, line 58: I think it's better when `LnurlAPIPrefix` and `LnurlDomain` are combined into a single variable (`LnurlDomain`).

All occurences in the code use `LnurlAPIPrefix` and `LnurlDomain` together (as `LnurlAPIPrefix + "." + LnurlDomain`) except this one:

``` golang
responseBody.Metadata = lnurlDescriptionHash(user.Nickname, controller.svc.Config.LnurlDomain)
```

and even there I think the full domain name should be used (including prefix).

lib/service/config.go
```diff
@@ -29,6 +31,9 @@ type Config struct {
 	MaxReceiveAmount      int64  `envconfig:"MAX_RECEIVE_AMOUNT" default:"0"`
 	MaxSendAmount         int64  `envconfig:"MAX_SEND_AMOUNT" default:"0"`
 	MaxAccountBalance     int64  `envconfig:"MAX_ACCOUNT_BALANCE" default:"0"`
+	LnurlDomain           string `envconfig:"LNURL_DOMAIN" default:"example.com"`
+	LnurlAPIPrefix        string `envconfig:"LNURL_API_PREFIX" default:"ln"`
```
prusnak, line 35: I think it's better when `LnurlAPIPrefix` and `LnurlDomain` are combined into a single variable (`LnurlDomain`).

All occurences in the code use `LnurlAPIPrefix` and `LnurlDomain` together (as `LnurlAPIPrefix + "." + LnurlDomain`) except this one:

``` golang
responseBody.Metadata = lnurlDescriptionHash(user.Nickname, controller.svc.Config.LnurlDomain)
```

and even there I think the full domain name should be used (including prefix).

lib/service/config.go
```diff
@@ -29,6 +31,9 @@ type Config struct {
...
+	SignedMessagePrefix   string `envconfig:"SIGNED_MESSAGE_PREFIX" default:"sign in into lndhub"`
```
prusnak, line 36: Is this used somewhere?

juligasa: No, removed.

prusnak:
There is a new function called `FindUserByLoginOrNickname` which looks for an user via `WHERE nickname = ? OR login = ? LIMIT 1`. This is problematic, since a new nickname would beat the previous login registration.

We should add a mechanism that disallows a user to set nickname that is either another user's login or another user's nickname.

Or is that what the `change_nickname_on_collision` + `ON CONFLICT` does? I am not really a PostgreSQL expert

juligasa:
> [There is a new function called FindUserByLoginOrNickname which looks for an user via WHERE nickname = ? OR login = ? LIMIT 1. This is problematic, since a new nickname would beat the previous login registration.](https://github.com/getAlby/lndhub.go/pull/217#issuecomment-1193314708)

The conflict resolver is meant to update the nickname automatically whenever it finds a collision on the [change_nickname_on_collision index](https://github.com/mintterteam/lndhub.go/blob/main/db/migrations/20220722171000_nicknames.up.sql#L9). This way, a client can call the /create endpoint with an existing login/password, and the nickname gets changed.
Nicknames [are unique](https://github.com/mintterteam/lndhub.go/blob/main/db/migrations/20220722171000_nicknames.up.sql#L5) so one user cannot override an old nickname. However, it may be this case:
```
lndhub=# select id, login, nickname from users where nickname = 'test2' or login = 'test2';
 id | login | nickname 
----+-------+----------
  8 | test2 | test3
  9 | test3 | test2

lndhub=# select id, login, nickname from users where nickname = 'test2' or login = 'test2' limit 1;
 id | login | nickname 
----+-------+----------
  8 | test2 | test3
  
lndhub=# select id, login, nickname from users where nickname = 'test3' or login = 'test3' limit 1;
 id | login | nickname 
----+-------+----------
  8 | test2 | test3
```
where login and nicknames are crossed. This could happen. In the case of postgres, the oldest takes precedence (like in your suggestion that login takes precedence). But maybe you are right it is better to make it explicit in code. However in the example above, user ID 9 is never found since its nickname is some else's login. I think the best solution would be disallow that a nickname can be someone else's login. Maybe when creating a user ??
```go
_, err := svc.FindUserByLogin(ctx, login)
if err == nil {
    return nil, fmt.Errorf("nickname already taken")
}
```

lib/service/config.go
juligasa: cherry-picked!

prusnak:
Last commit introduced SignedMessagePrefix by accident

lib/security/password.go
juligasa: One of my teammates raised some concerns about breaking the de-facto standard for storing passwords in databases. I think I can revert it to `bcrypt` and handle the password,login collisions on code instead of doing it at the db level. I am doing it anyway for enforcing login/nickname uniqueness, so I don't think is a step back and at least we don't weaken the security we previously had.

kiwiidb:
Hey @juligasa , thank you for submitting this PR, great that you guys are also working on LNDhub.go 💯 

However, after reviewing this PR (sorry that it took so long), we (@bumi and myself) are of the opinion that the proposed features are out of scope of the LNDhub.go project. We are wary of feature creep and feel like LNDhub.go should be an accounting system and a concise middleware between a Lightning node and an account holder, with an API that allows it to interact with supporting services as part of a broader stack of microservices, if desired. 

The features proposed in this PR (nicknames, LNURLP endpoints, signature auth for creating accounts) should be implemented as external services that interact with LNDhub. For example, to only allow creating accounts with some signature authorization scheme, you can block off access to the endpoint on an ingress level, have a 2nd service act as a proxy between the outside world and LNDhub.go, and route the `/create` requests through that proxy. LN Addresses is also something that we at Alby have deliberately implemented as a seperate service that interacts with LNDhub.go, instead of implementing it in LNDhub.go itself.

The feature of configuring macaroon/cert to be read from files has already been merged from the seperate PR that was made by @prusnak .

juligasa:
> Hey @juligasa , thank you for submitting this PR, great that you guys are also working on LNDhub.go 100
> 
> However, after reviewing this PR (sorry that it took so long), we (@bumi and myself) are of the opinion that...
Sure! That was initially the intention of this PR. Just share a bunch of changes we did to cover our necessities and maybe some of them could be extracted to their own PR's.

integration_tests/lnd_mock.go
```diff
@@ -101,7 +101,7 @@ func (mlnd *MockLND) AddInvoice(ctx context.Context, req *lnrpc.Invoice, options
 	copy(invoice.PaymentAddr[:], req.PaymentAddr)
 	if len(req.DescriptionHash) != 0 {
 		invoice.DescriptionHash = &[32]byte{}
-		copy(req.DescriptionHash, invoice.DescriptionHash[:])
```
juligasa, line 104: Wrong src,dst

integration_tests/lnd_mock.go
```diff
@@ -227,15 +227,16 @@ func (mlnd *MockLND) DecodeBolt11(ctx context.Context, bolt11 string, options ..
...
 		RouteHints:  []*lnrpc.RouteHint{},
 		PaymentAddr: []byte{},
 		NumMsat:     int64(*inv.MilliSat),
 		Features:    map[uint32]*lnrpc.Feature{},
 	}
 	if inv.DescriptionHash != nil {
-		result.DescriptionHash = string(inv.DescriptionHash[:])
+		result.DescriptionHash = hex.EncodeToString(inv.DescriptionHash[:])
+	} else if inv.Description != nil { // according to spec if DescriptionHash is present description must not be present
+		result.Description = *inv.Description
```
juligasa, line 239: Spec compliance

juligasa (commented):
Maybe it's worth stripping lnd_mock.go changes into a PR to patch the bugs there

PR #218, opened 2022-07-22
macaroon and cert can be provided as files

prusnak:
Extracted from https://github.com/getAlby/lndhub.go/pull/217


bumi (commented):
utACK

kiwiidb (approved):
Looks good, thanks

PR #219, opened 2022-07-28
introduce config.AllowCustomLogin and config.AllowCustomPassword

prusnak:
These two options enable/disable the possibility to request a particular login/password while creating a new user.

ISTM it's quite dangerous to allow users to specify their own login/password and I'd prefer not having this option for my instance of LndHub.go.

🤔 not sure why the tests fail with `custom login is not allowed`. I tried to do the changes in a way the default behaviour does not change at all, so I am pretty much puzzled about this.


prusnak:
superseded by https://github.com/getAlby/lndhub.go/pull/220

PR #220, opened 2022-07-29
Introduce config.MinPasswordEntropy option

prusnak:
that allows an operator to set minimum entropy (in bits) of a password to be accepted during account creation


kiwiidb (approved):
ACK

PR #224, opened 2022-08-08
don't omit payment error when empty

### LINKED ISSUE #223: Make /payinvoice compatible with Zeus
We are not returning the right response according to https://github.com/ZeusLN/zeus/blob/master/stores/TransactionsStore.ts#L286. The `payment_error` should be an empty string, and we are not sending anything right now.

kiwiidb:
Fixes #223 , testing on testnet when build is ready.


bumi:
so this then always returns an empty string if no PaymentError is set, correct?

bumi (approved):

prusnak (approved):

kiwiidb:
> so this then always returns an empty string if no PaymentError is set, correct?

yes

PR #231, opened 2022-09-27
Look up status of outgoing payments at startup

kiwiidb:
#232 
At startup, do the following things before starting the main server:

- look up all payments that are `outgoing` AND `initialized`
- For each one, start a goroutine that will subscribe to LND's `routerrpc` for the status of the payment.
- There should be 3 possibilities for the status: an error (not found), `SUCCEEDED`, `FAILED` (we don't want `IN_FLIGHT` updates).
-Handle succesful or failed payments once the payments are completed.

This PR is now ready for review, but this afternoon I will add integration tests for the following 2 scenarios:

- Payment succesful after LNDhub startup
- Payment failed after LNDhub startup


lib/service/checkpayments.go
```diff
@@ -62,18 +63,24 @@ func (svc *LndhubService) TrackOutgoingPaymentstatus(ctx context.Context, invoic
 			return
 		}
 		if payment.Status == lnrpc.Payment_FAILED {
-			svc.Logger.Infof("Failed payment detected: %v", payment)
+			svc.Logger.Infof("Failed payment detected: hash %s, reason %s", payment.PaymentHash, payment.FailureReason)
 			//todo handle failed payment
-			//return svc.HandleFailedPayment(ctx, invoice, entry, fmt.Errorf(payment.FailureReason.String()))
+			err = svc.HandleFailedPayment(ctx, invoice, entry, fmt.Errorf(payment.FailureReason.String()))
+			if err != nil {
+				svc.Logger.Errorf("Error handling failed payment %v: %s", invoice, err.Error())
```
bumi, line 73: this sounds like an error that needs attention? maybe this should be a sentry notification?

lib/service/checkpayments.go
```diff
@@ -62,18 +63,24 @@ func (svc *LndhubService) TrackOutgoingPaymentstatus(ctx context.Context, invoic
...
 			invoice.Fee = payment.FeeSat
 			invoice.Preimage = payment.PaymentPreimage
-			svc.Logger.Infof("Completed payment detected: %v", payment)
-			//todo handle succesful payment
-			//return svc.HandleSuccessfulPayment(ctx, invoice, entry)
-			return
+			svc.Logger.Infof("Completed payment detected: hash %s", payment.PaymentHash)
+			err = svc.HandleSuccessfulPayment(ctx, invoice, entry)
+			if err != nil {
+				svc.Logger.Errorf("Error handling successful payment %v: %s", invoice, err.Error())
```
bumi, line 85: same here, this sounds like we should be notified of this?

lib/service/checkpayments.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/getAlby/lndhub.go/db/models"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
+)
+
+func (svc *LndhubService) CheckAllPendingOutgoingPayments(ctx context.Context) (err error) {
+	//check database for all pending payments
+	pendingPayments := []models.Invoice{}
+	//since this part is synchronously executed before the main server starts, we should not get into race conditions
+	err = svc.DB.NewSelect().Model(&pendingPayments).Where("state = 'initialized'").Where("type = 'outgoing'").Scan(ctx)
```
bumi, line 17: should we add a time limit here? even though after this change we should not have old initialized ones, correct?

lib/service/checkpayments.go
```diff
@@ -0,0 +1,94 @@
...
+			err = svc.HandleSuccessfulPayment(ctx, invoice, entry)
+			if err != nil {
+				svc.Logger.Errorf("Error handling successful payment %v: %s", invoice, err.Error())
+				return
+			}
+			svc.Logger.Infof("Updated completed payment: hash %s", payment.PaymentHash)
+			return
+		}
+		//Since we shouldn't get in-flight updates we shouldn't get here
+		svc.Logger.Warnf("Got an unexpected in-flight update %v", payment)
```
bumi, line 92: as it is unexpected should we use sentry here?

PR #235, opened 2022-10-03
set fee reserve to 0 if destination is us

### LINKED ISSUE #233: improve fee reserve check
Payments to internal accounts should have a 0 fee reserve requirement.

kiwiidb:
Closes #233 
In case of an internal payment, the fee reserve should be 0.
This change allow clients to create a "Drain account" UX by combining different calls:
- make a payment (or more than 1) leaving just the fee reserve in the account, using a Lightning Address or 0 amount invoice.
- send the rest of the funds to some internal wallet (eg. hello@getalby.com or someone else).


bumi (commented):
can we add a test for this?

kiwiidb:
Yes, done.

bumi (approved):

PR #241, opened 2022-12-05
add multikeysend endpoint

### LINKED ISSUE #230: Provide API to send multiple split payments with one request
A common case in the Podcasting 2.0 world is "split payments". That means the podcaster defines multiple payment recipients and the clients send a payment to each of those. 

Right now clients need to gather payment details and loop through those and send requests for each payment. 

We could provide an API to send multiple payments with one HTTP call. This would lower the complexity for the clients and lower the possibility to be overloaded with too many more or less parallel requests for the same account.

kiwiidb:
This PR fixes #230
Request:

Screenshot 2022-12-06 at 15 10 02

Response:
Screenshot 2022-12-06 at 15 10 30

The response can contain both succesful and failed keysend payments.


bumi (approved):
looks good to me. 
did not run it though.

PR #246, opened 2022-12-09
fix http status header

kiwiidb:
Fixes an error that must have slipped through our rigorous review process.

### DIFF
--- a/controllers_v2/keysend.ctrl.go
+++ b/controllers_v2/keysend.ctrl.go
@@ -78,7 +78,7 @@ func (controller *KeySendController) KeySend(c echo.Context) error {
 	result, err := controller.SingleKeySend(c, &reqBody, userID)
 	if err != nil {
 		c.Logger().Errorf("Failed to send keysend: %s", err.Message)
-		return c.JSON(err.Code, err)
+		return c.JSON(http.StatusInternalServerError, err)
 	}
 	return c.JSON(http.StatusOK, result)
 }


controllers_v2/keysend.ctrl.go
```diff
@@ -78,7 +78,7 @@ func (controller *KeySendController) KeySend(c echo.Context) error {
 	result, err := controller.SingleKeySend(c, &reqBody, userID)
 	if err != nil {
 		c.Logger().Errorf("Failed to send keysend: %s", err.Message)
-		return c.JSON(err.Code, err)
+		return c.JSON(http.StatusInternalServerError, err)
```
bumi, line 81: when does this fail?

kiwiidb: See lndhub-notifications on slack, we're sending `2` as http status header

bumi: yes, understood that. but why did the keysend fail? 

wondering if a 500 error is correct? 
and we probably should return a valid JSON response in the same format as others: https://github.com/getAlby/lndhub.go/blob/main/lib/responses/errors.go

kiwiidb: keysend fails because the user didn't have enough money.

kiwiidb: We are returning a valid JSON response. "err" is not an error (not the greatest var name here) it's a struct.

bumi: if it is not enough money then it should be this error: https://github.com/getAlby/lndhub.go/blob/main/lib/responses/errors.go#L35

bumi: if some server error then this: https://github.com/getAlby/lndhub.go/blob/main/lib/responses/errors.go#L17

bumi (approved):

PR #247, opened 2022-12-14
GRPC server

### LINKED ISSUE #248: GRPC server
To allow multiple downstream services to subscribe to events about new invoices or payments.

See discussion at https://github.com/getAlby/meta/issues/203

kiwiidb:
Fixes #248 

Based on https://grpc.io/docs/languages/go/basics/

- I've used https://json-to-proto.github.io/ to convert the webhook payload to a protobuf definition. Since you cannot have arbitrary keys in protobuf I've changed the custom records from a map to an array.
- Used protoc to generate the files in the "lndhubrpc" package. 
- In case the client needs invoices from a certain id onwards we fetch those first and then return the real-time settled invoices. To avoid race conditions where we send the same invoice twice (an invoice is settled while we fetch the invoices from the database), we check the index of the last invoice from the database array against the id from the invoices of the channel. I'm not sure if this is necessary though.
- Copied the webhook test and changed some things to refactor it to a grpc test.


bumi:
@jankoegel tagging you here, fyi. We will need to write a GRPC consumer in ruby.

kiwiidb:
> @jankoegel tagging you here, fyi. We will need to write a GRPC consumer in ruby.

Client issue is https://github.com/getAlby/getalby.com/issues/286 btw. I can also do it, doesn't seem to be much work.

I would make it seperate from the main rails app like we currently also do with the cron job, but this app would then just be constantly on consume the gRPC stream, and update the database.

PR #251, opened 2022-12-23
respond with 500 if no payment sent

### LINKED ISSUE #250: multikeysend: return error in case no payments completed
We should check the amount of payments that have actually completed. If it's 0, don't return 200 status code.

kiwiidb:
Fixes #250

### DIFF
--- a/controllers_v2/keysend.ctrl.go
+++ b/controllers_v2/keysend.ctrl.go
@@ -109,6 +109,7 @@ func (controller *KeySendController) MultiKeySend(c echo.Context) error {
 	result := &MultiKeySendResponseBody{
 		Keysends: []KeySendResult{},
 	}
+	singleSuccesfulPayment := false
 	for _, keysend := range reqBody.Keysends {
 		keysend := keysend
 		res, err := controller.SingleKeySend(c, &keysend, userID)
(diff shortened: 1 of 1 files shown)


controllers_v2/keysend.ctrl.go
```diff
@@ -125,8 +126,13 @@ func (controller *KeySendController) MultiKeySend(c echo.Context) error {
 		result.Keysends = append(result.Keysends, KeySendResult{
 			Keysend: res,
 		})
+		singleSuccesfulPayment = true
 	}
-	return c.JSON(http.StatusOK, result)
+	status := http.StatusOK
+	if !singleSuccesfulPayment {
+		status = http.StatusInternalServerError
```
bumi, line 133: is this a correct HTTP status code?
internal server error indicates it's an exception an unexpected bug in our system

payments might fail because the recipient can not receive anything. 

maybe a 422 - Unprocessable Entity?

kiwiidb: That would deserve it's own PR as we would need to change it everywhere we're making a payment. I'd rather have it consistent for now.

bumi: we return a 500 if a payment fails? damn, also not a good http cititzen.

PR #253, opened 2022-12-26
Add DB indexes migration

bumi:


db/migrations/20221226203000_db_index.up.sql
```diff
@@ -0,0 +1,59 @@
...
+
+CREATE INDEX IF NOT EXISTS index_invoices_on_expires_at ON invoices(expires_at)
+
+--migration:split
+
+CREATE INDEX IF NOT EXISTS index_invoices_on_settled_at ON invoices(settled_at)
+
+--migration:split
+
+CREATE INDEX IF NOT EXISTS index_invoices_on_type_and_r_hash_and_state ON invoices(type, r_hash, state);
```
reneaaron, line 33: This one should speed up query 1 from here:

https://github.com/getAlby/meta/issues/290

db/migrations/20221226203000_db_index.up.sql
```diff
@@ -0,0 +1,59 @@
...
+CREATE INDEX IF NOT EXISTS index_transaction_entries_on_debit_account_id ON transaction_entries(debit_account_id);
+
+--migration:split
+
+
+CREATE INDEX IF NOT EXISTS index_accounts_user_id ON accounts(user_id);
+
+--migration:split
+
+CREATE INDEX IF NOT EXISTS index_accounts_type ON accounts(type);
```
reneaaron, line 59: Query 3 is missing: 

SELECT "invoices".* FROM "invoices" WHERE "invoices"."user_id" = $1 AND "invoices"."invoice_type" = $2 AND "invoices"."state" = $3 ORDER BY "invoices"."created_at" DESC LIMIT $4 OFFSET $5

➡️ ```CREATE INDEX IF NOT EXISTS index_invoices_on_userid_and_invoicetype_and_state_and_created_at ON invoices(user_id, invoice_type, state, created_at DESC);```

bumi: those are not from lndhub.

PR #264, opened 2022-12-28
feature: return custom records

### LINKED ISSUE #261: Multikeysend: return customkey/customvalue
In order to know which keysend payment failed exactly, we must return these values as well.

kiwiidb:
Fixes #261 
Tested locally

### DIFF
--- a/controllers_v2/keysend.ctrl.go
+++ b/controllers_v2/keysend.ctrl.go
@@ -46,7 +46,7 @@ type KeySendResponseBody struct {
 	Description     string            `json:"description,omitempty"`
 	DescriptionHash string            `json:"description_hash,omitempty"`
 	Destination     string            `json:"destination,omitempty"`
-	CustomRecords   map[string]string `json:"custom_ecords" validate:"omitempty"`
+	CustomRecords   map[string]string `json:"custom_records" validate:"omitempty"`
 	PaymentPreimage string            `json:"payment_preimage,omitempty"`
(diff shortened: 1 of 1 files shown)


controllers_v2/keysend.ctrl.go
```diff
@@ -41,13 +41,14 @@ type KeySendResult struct {
...
-	DescriptionHash string `json:"description_hash,omitempty"`
-	Destination     string `json:"destination,omitempty"`
-	PaymentPreimage string `json:"payment_preimage,omitempty"`
-	PaymentHash     string `json:"payment_hash,omitempty"`
+	Amount          int64             `json:"amount"`
+	Fee             int64             `json:"fee"`
+	Description     string            `json:"description,omitempty"`
+	DescriptionHash string            `json:"description_hash,omitempty"`
+	Destination     string            `json:"destination,omitempty"`
+	CustomRecords   map[string]string `json:"customRecords" validate:"omitempty"`
```
bumi, line 49: damn, just realizing we have camel case here but snake case for the other attributes :(

kiwiidb: yeah that sucks.

bumi: shall we add `custom_records` also? then we can deprecate the CustomRecords in the future?

kiwiidb: Here we can just add the snake case as the only thing since this is a totally new field anyway. It's the request body that is more tricky.

PR #277, opened 2023-01-06
invoice: lookup more performant by hash

### LINKED ISSUE #271: Investigate net.Operror: i/o timeout
Internal payments are failing because of this error:

https://sentry.io/organizations/nomadiclabs/issues/3625901873
Things to try
- [x] reproduce on dev environment
- [ ] write intentionally slow query and cancel context?
- [ ] Run with BUNDEBUG in production

kiwiidb:
Possible fix #271


bumi (approved):
can't think of any issue with the rhash - it is always there an query able.

PR #280, opened 2023-01-09
more accurate http status codes for keysend

### LINKED ISSUE #270: Better error messages on internal keysend payments
Currently we return a HTTP 500 with some error "result not found" when an internal keysend payment is sent without providing any account identifier. (customkey).

We should return a specific error message saying that no account details are provided (customKey/Value) is required. 

likely this is the error here:
https://github.com/getAlby/lndhub.go/blob/de592ffccc00164cf61b74b7678b898e0c671ae5/lib/service/invoicesubscription.go#L25

kiwiidb:
Fixes #270 
Tested locally. 

```
➜  Alby http POST localhost:3000/v2/payments/keysend < keysend.json Authorization:"Bearer $token"
HTTP/1.1 400 Bad Request
Content-Length: 110
Content-Type: application/json; charset=UTF-8
Date: Mon, 09 Jan 2023 14:02:42 GMT
X-Request-Id: vddq9Yq3JPt5gYj5QSGk91cpcL4NYWye

{
    "code": 8,
    "error": true,
    "message": "Internal keysend payments require the custom record 696969 to be present."
}
```

Other endpoints can also be updated to use the status code from `ErrorResponse` to improve the accuracy of the http status code returned. However this should be for another PR.

### DIFF
--- a/controllers/keysend.ctrl.go
+++ b/controllers/keysend.ctrl.go
@@ -70,6 +70,15 @@ func (controller *KeySendController) KeySend(c echo.Context) error {
 		}
 	}
 
+	if reqBody.Destination == controller.svc.IdentityPubkey && reqBody.CustomRecords[strconv.Itoa(service.TLV_WALLET_ID)] == "" {
+		return c.JSON(http.StatusBadRequest, &responses.ErrorResponse{
+			Error:          true,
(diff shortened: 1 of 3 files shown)


bumi (approved):
utACK

PR #281, opened 2023-01-09
allow 0 amount invoices in v2 api

### LINKED ISSUE #242: Creating invoices with `amount: 0` is not possible
It doesn't seem to be possible to create an invoice via the `POST /v2/invoices` endpoint with the following body:

```
{
  "amount": 0,
  "description: "whatever"
}
```

The error log on the server is:

```
{"level":"error","id":"(...)","time":"(...)","message":"Invalid addinvoice request body: Key: 'AddInvoiceRequestBody.Amount' Error:Field validation for 'Amount' failed on the 'required' tag"}
```

The [validation definition](https://github.com/getAlby/lndhub.go/blob/1efba841907bcca2cf63d6057b12078ff065309f/controllers_v2/invoice.ctrl.go#L128) is `"required,gte=0"`. It looks like `0` was originally meant to be a valid value (hence `gte=0`). However, after some quick googling, it seems like this is not compatible with `required` which, [according to the docs](https://pkg.go.dev/gopkg.in/go-playground/validator.v9#hdr-Required), "(...) ensures value is not zero".

I'm not a Go expert so I don't have a solution at hand but it seems like we'd need a different validation if we want to allow 0-value invoices (which would be a nice feature, imo). Not sure if I'll manage to open a PR myself anytime soon so I figured I'd just report it here. Let me know what you think!

kiwiidb:
Fixes #242

### DIFF
--- a/controllers_v2/invoice.ctrl.go
+++ b/controllers_v2/invoice.ctrl.go
@@ -125,7 +125,7 @@ func (controller *InvoiceController) GetIncomingInvoices(c echo.Context) error {
 }
 
 type AddInvoiceRequestBody struct {
-	Amount          int64  `json:"amount" validate:"required,gte=0"`
+	Amount          int64  `json:"amount" validate:"gte=0"`
 	Description     string `json:"description"`
 	DescriptionHash string `json:"description_hash" validate:"omitempty,hexadecimal,len=64"`
 }
--- a/integration_tests/expected_requests_and_responses.go
+++ b/integration_tests/expected_requests_and_responses.go
@@ -28,6 +28,11 @@ type ExpectedAddInvoiceRequestBody struct {
 	Memo            string      `json:"memo"`
 	DescriptionHash string      `json:"description_hash" validate:"omitempty,hexadecimal,len=64"`
 }
+type ExpectedV2AddInvoiceRequestBody struct {
+	Amount          int64  `json:"amount"` // amount in Satoshi
+	Memo            string `json:"description"`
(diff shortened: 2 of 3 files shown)


bumi:
should we add a test for this?

kiwiidb:
Unit test added, proof that it fails when the fix is not applied:

```
➜  integration_tests git:(feature/zero-amount-invoices) ✗ go test -run InvoiceSuite/TestZeroAmtTestSuite 
{"level":"info","time":"2023-01-12T18:00:32+01:00","message":"Starting invoice subscription from index: 0"}
{"time":"2023-01-12T18:00:32.71847+01:00","level":"ERROR","prefix":"echo","file":"invoice.ctrl.go","line":"161","message":"Invalid addinvoice request body: Key: 'AddInvoiceRequestBody.Amount' Error:Field validation for 'Amount' failed on the 'required' tag"}
--- FAIL: TestInvoiceSuite (0.15s)
    --- FAIL: TestInvoiceSuite/TestZeroAmtTestSuite (0.01s)
        invoice_test.go:71: 
                Error Trace:    /Users/kwinten/Alby/lndhub.go/integration_tests/invoice_test.go:71
                Error:          Not equal: 
                                expected: 200
                                actual  : 400
                Test:           TestInvoiceSuite/TestZeroAmtTestSuite
FAIL
exit status 1
FAIL    github.com/getAlby/lndhub.go/integration_tests  0.597s
```

bumi (approved):

PR #288, opened 2023-01-13
atomic failed payment handling

### LINKED ISSUE #286: Handle failed invoice: db transactions need to be atomic.
Currently it is possible for a tx entry to be added and then (due to network or db issues) for the invoice update to fail, which can lead to inconsistencies. https://github.com/getAlby/lndhub.go/blob/main/lib/service/invoices.go#L258-L259 needs to be a single db transaction.

kiwiidb:
Fixes #286 
This brings handling a failed payment in line with the way we add incoming payments to a user's account.

### DIFF
--- a/lib/service/invoices.go
+++ b/lib/service/invoices.go
@@ -3,6 +3,7 @@ package service
 import (
 	"context"
 	"crypto/sha256"
+	"database/sql"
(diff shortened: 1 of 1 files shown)


bumi (approved):
lgtm

PR #291, opened 2023-01-17
Feature/graceful shutdown

### LINKED ISSUE #285: Graceful shutdown of all goroutines
Currently, only the server is shut down gracefully, but all goroutines should be able to handle a graceful shutdown.

kiwiidb:
Fixes #285 
Based on this [article](https://www.rudderstack.com/blog/implementing-graceful-shutdown-in-go/).

We need to shutdown a maximum of 3 servers:
- The main Echo REST server
- (optional) The Prometheus metrics server
- (optional) The gRPC server
These servers are shut down gracefully by calling some kind of `Stop` or `Shutdown` function from `main.go` after `backgroundCtx` is canceled, which is done when the signal receives an `os.Interupt` (Ctrl-c) signal.

Other than that there are the background routines:
- Incoming invoice gRPC subscription routine
- Payment tracking routines
- Webhook dispatching routines

These are also canceled by canceling the `backgroundCtx` context. The waitgroup waits until all processes are finished before the program is exited.

### DIFF
--- a/lib/service/invoicesubscription.go
+++ b/lib/service/invoicesubscription.go
@@ -243,11 +243,6 @@ func (svc *LndhubService) InvoiceUpdateSubscription(ctx context.Context) error {
 			if err != nil {
 				svc.Logger.Errorf("Error processing invoice update subscription: %v", err)
 				sentry.CaptureException(err)
-				// TODO: close the stream somehoe before retrying?
-				// Wait 30 seconds and try to reconnect
-				// TODO: implement some backoff
-				time.Sleep(30 * time.Second)
-				invoiceSubscriptionStream, _ = svc.ConnectInvoiceSubscription(ctx)
 				continue
 			}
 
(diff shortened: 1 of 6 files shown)


lib/service/invoicesubscription.go
```diff
@@ -243,11 +243,6 @@ func (svc *LndhubService) InvoiceUpdateSubscription(ctx context.Context) error {
 			if err != nil {
 				svc.Logger.Errorf("Error processing invoice update subscription: %v", err)
 				sentry.CaptureException(err)
-				// TODO: close the stream somehoe before retrying?
```
kiwiidb, line 246: I needed to remove this because the "error" that is returned here, if there is one, is most likely the context cancellation. Even if it is some kind of other error, I feel it doesn't really make sense to add these kind of ad-hoc reconnects here.

LucasRouckhout (approved):

PR #293, opened 2023-01-17
update dependencies

### LINKED ISSUE #276: Bun upgrade causes migrations to fail
See #275 

Figure out how we can upgrade bun without breaking the migrations.

kiwiidb:
Fixes #276 

Just ran a `go get -u`, which upgrades all our dependencies to the latest available version.
Supersedes all the different dependabot PR's.


bumi:
I will test this because I think we need to update some migration files maybe.

bumi (approved):
looks good to me. 
Migrations work well, too.

PR #300, opened 2023-02-04
Use rabbitmq as an alternative to publish invoice events 

### LINKED ISSUE #299: Feature: RabbitMQ publisher
We want to publish invoice events to RabbitMQ as a replacement of the current webhook-based event publishing. 
The routine should be completely analogous to the current webhook dispatching routine. As a routing key for the topic exchange we could use `<type>.<state>.invoices`, where `type` can be `incoming` our `outgoing`, and `state` can be `error` or `settled`. Error events can be supported in the future (previously we didn't dispatch them over the webhook). The payload should be the same as the webhook payload.

LucasRouckhout:
fixes #299


integration_tests/rabbitmq_test.go
```diff
@@ -0,0 +1,128 @@
...
+		false,
+		false,
+		false,
+		nil,
+	)
+	assert.NoError(suite.T(), err)
+
+	msg := <-m
+
+	var recievedInvoice models.Invoice
```
kiwiidb, line 102: typo

PR #305, opened 2023-02-13
feat: update invoice expiration date to one week

### LINKED ISSUE #304: Increase invoice expiration
Current: 1 day
Proposal: 1 week

reneaaron:
Fixes #304


bumi:
can you describe why? (best actually in the commit message)

kiwiidb:
In any case I would rather like to see this as part of the config, with a default that is the same as what we want to run in production after we've agreed on it.

reneaaron:
There are also other aspects to the invoice expiration that I didn't consider. (the expiration window allows you to limit exchange rate risks, e.g. when you send an invoice to a client) 

I don't see an immediate reason to change something here.

PR #310, opened 2023-02-16
feature: add db pool config

### LINKED ISSUE #307: DB connection pool configuration
https://making.pusher.com/production-ready-connection-pooling-in-go/

TL;DR:
Go doesn’t seem to have good defaults for production environments, so you should check the defaults.
I think we could raise those defaults to some numbers that make sense for us and see if that helps?
Some numbers to get the discussion started:
- SetMaxOpenConns  - 10 (we have 99 available on db level, right?)
- SetMaxIdleConns  - 5 (50% of the max open conns)
- SetConnMaxLifetime  - 30 min

kiwiidb:
Fixes #307

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -4,19 +4,25 @@ import (
 	"database/sql"
 	"fmt"
 	"strings"
+	"time"
 
+	"github.com/getAlby/lndhub.go/lib/service"
 	"github.com/uptrace/bun"
 	"github.com/uptrace/bun/dialect/pgdialect"
 	"github.com/uptrace/bun/driver/pgdriver"
 	"github.com/uptrace/bun/extra/bundebug"
 )
 
-func Open(dsn string) (*bun.DB, error) {
+func Open(config *service.Config) (*bun.DB, error) {
 	var db *bun.DB
+	dsn := config.DatabaseUri
 	switch {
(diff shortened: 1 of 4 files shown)


db/db.go
```diff
@@ -4,19 +4,25 @@ import (
...
+func Open(config *service.Config) (*bun.DB, error) {
 	var db *bun.DB
+	dsn := config.DatabaseUri
 	switch {
 	case strings.HasPrefix(dsn, "postgres://") || strings.HasPrefix(dsn, "postgresql://") || strings.HasPrefix(dsn, "unix://"):
 		dbConn := sql.OpenDB(pgdriver.NewConnector(pgdriver.WithDSN(dsn)))
 		db = bun.NewDB(dbConn, pgdialect.New())
+		db.SetMaxOpenConns(config.DatabaseMaxConns)
+		db.SetMaxIdleConns(config.DatabaseMaxIdleConns)
+		db.SetConnMaxLifetime(time.Duration(config.DatabaseConnMaxLifetime) * time.Second)
```
bumi, line 25: that should be `time.Minute` for 30 miniutes - and not seconds?

kiwiidb: I've updated the default to 1800 - I would prefer seconds as a default time unit.

bumi (commented):
utack

PR #316, opened 2023-02-17
Invoice subscription: safety improvements

### LINKED ISSUE #312: Invoice subscription: fail on error
If [this call](https://github.com/getAlby/lndhub.go/blob/catch-add-index-error/lib/service/invoicesubscription.go#L246) fails, it is better to crash because then we can try to reconnect. Just `continue`ing will lead to a very high error rate if the issue isn't fixed.

### LINKED ISSUE #313: Invoice subscription: fail on db error
If [this call](https://github.com/getAlby/lndhub.go/blob/catch-add-index-error/lib/service/invoicesubscription.go#L222) fails, then we will not ask LND for older invoices, which means that there are potentially paid invoices that were not credited. This error is serious enough to restart LNDhub.

kiwiidb:
Fixes #312 , fixes #313 
This PR changes the behaviour of the invoice subscription routine in case of an error:
- If the database call at startup time fails, we return an error which will crash LNDhub and force a restart. 
- If the `Recv()` call fails (probably because LND is down), then we will also force LNDhub to restart, so it can try to reconnect to LND.

 Other than that
- It introduces a database timeout config option, with a default value of 1 minutes (up from 5 seconds)
- It introduces a safety buffer of 14 hours to make sure we absolutely did not miss any invoices which might have been paid while LNDhub was down.

### DIFF
--- a/lib/service/invoicesubscription.go
+++ b/lib/service/invoicesubscription.go
@@ -222,6 +222,8 @@ func (svc *LndhubService) ConnectInvoiceSubscription(ctx context.Context) (lnd.S
 	// IF we found an invoice we use that index to start the subscription
 	if err == nil {
 		invoiceSubscriptionOptions = lnrpc.InvoiceSubscription{AddIndex: invoice.AddIndex - 1} // -1 because we want updates for that invoice already
+	} else {
+		sentry.CaptureException(err)
 	}
 	svc.Logger.Infof("Starting invoice subscription from index: %v", invoiceSubscriptionOptions.AddIndex)
 	return svc.LndClient.SubscribeInvoices(ctx, &invoiceSubscriptionOptions)
--- a/lib/service/invoicesubscription.go
+++ b/lib/service/invoicesubscription.go
@@ -217,12 +217,14 @@ func (svc *LndhubService) ConnectInvoiceSubscription(ctx context.Context) (lnd.S
 	var invoice models.Invoice
 	invoiceSubscriptionOptions := lnrpc.InvoiceSubscription{}
 	// Find the oldest NOT settled AND NOT expired invoice with an add_index
-	//  Note: expired invoices will not be settled anymore, so we don't care about those
-	err := svc.DB.NewSelect().Model(&invoice).Where("invoice.settled_at IS NULL AND invoice.add_index IS NOT NULL AND invoice.expires_at >= now()").OrderExpr("invoice.id ASC").Limit(1).Scan(ctx)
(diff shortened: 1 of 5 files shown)


LucasRouckhout (approved):

reneaaron:
@kiwiidb Should we also add an index for the invoice subscription query, too? It's only a matter of time until this query takes longer than the timeout.

`explain analyze select * from public.invoices where invoices.settled_at IS NULL AND invoices.add_index IS NOT NULL AND invoices.expires_at >= now() order by invoices.id ASC` 

leads to some really awful query plan as there are no indexes to be used currently. 

I think we would need to add a combined column index on the used columns to speed this up.

lib/service/invoicesubscription.go
```diff
@@ -236,14 +243,16 @@ func (svc *LndhubService) InvoiceUpdateSubscription(ctx context.Context) error {
 	for {
 		select {
 		case <-ctx.Done():
-			return fmt.Errorf("Context was canceled")
+			return context.Canceled
```
bumi, line 246: @kiwiidb this is a differnt/correct type of error, correct?

lib/service/invoicesubscription.go
```diff
@@ -236,14 +243,16 @@ func (svc *LndhubService) InvoiceUpdateSubscription(ctx context.Context) error {
...
 		default:
 			// receive the next invoice update
 			rawInvoice, err := invoiceSubscriptionStream.Recv()
+			// in case of an error, we want to return and restart LNDhub
+			// in order to try and reconnect the gRPC subscription
 			if err != nil {
 				svc.Logger.Errorf("Error processing invoice update subscription: %v", err)
 				sentry.CaptureException(err)
-				continue
+				return err
```
bumi, line 255: where will this error end up?

bumi:
@reneaaron of course :) indexes need to be pimped

lib/service/invoicesubscription.go
kiwiidb: it's just a slightly different string (the standard one from the context lib)

lib/service/invoicesubscription.go
kiwiidb: [here](https://github.com/getAlby/lndhub.go/blob/catch-add-index-error/main.go#L184)

bumi: ah because it is returning, otherwise it would loop and not reach there, ok. thanks.

bumi (approved):

PR #320, opened 2023-02-18
Buffered PubSub channels

LucasRouckhout:
# The issue

The PubSub implementation for invoices in lndhub is heavily inspired by the one described [here](https://eli.thegreenplace.net/2020/pubsub-using-channels-in-go/).

The use of unbuffered channels in the lndhub PubSub system have the potential to cause cascading latency spikes throughout the system. This is indeed mentioned in the above mentioned blog.

>The critical issue here is blocking. Recall the sending loop in Publish:

```golang
for _, ch := range ps.subs[topic] {
  ch <- msg
}
```
>If ch is unbuffered, then ch <- msg will block until the message is consumed by a receiver. This prevents Pubsub from notifying other subscribers on the same channel. Is this the desired behavior? Not likely. Unless you can guarantee that receivers consume messages from subscriptions very quickly, it may be a good idea to buffer the channels. A buffer of size 1 would make it much more robust, wherein the publishing loop could finish notifying all topic subscribers quickly (unless a receiver is badly backed up and hasn't even consumed the previous message yet).

Sending or receiving from an unbuffered channel blocks the running thread. Solution, make the channel buffered ones to provide some breathing room for potential latency spikes. To make sure no accidental unbuffered channel is introduces in a later change we should actually create the channels during the call to `Subscribe`.


kiwiidb (approved):
Looks good, just fixed conflicts.

Issue #328, opened 2023-02-25
Consume lnrpc invoices from RabbitMQ topic


kiwiidb:
Part of the migration to an event-driven architecture.
Consuming from RabbitMQ would give several benefits including:
- Being able to receive invoice updates from multiple Lightning nodes
- LNDhub being able to (partially) work with the Lightning node temporarily offline

PR #329, opened 2023-02-25
Use the `lnd_invoice` exchange instead of a GRPC subscription to LND

LucasRouckhout:
#328


LucasRouckhout:
Starting from new branch with new approach

PR #333, opened 2023-03-05
Use rabbitmq as the consumer for LND invoices instead of gRPC subscription

LucasRouckhout:
Includes a refactor with regards to rabbitmq as well.


kiwiidb (approved):
Paired on this with Lucas today, ready to be tested on the testnet cluster.

PR #346, opened 2023-04-05
remove grpc server

kiwiidb:
It was a mistake, it adds too much complexity. We will use the admin API key to implement admin functions.


bumi:
👍

PR #347, opened 2023-04-05
remove ws

kiwiidb:
Nobody uses them (we have 1 hit in 2 months).
They won't work anymore when we deploy 2 lndhub replicas.


bumi:
👍

PR #350, opened 2023-04-18
Feature/republish invoices

kiwiidb:


republish_invoices/main.go
```diff
@@ -0,0 +1,86 @@
...
+		logrus.Infof("Publishing invoice with hash %s", inv.RHash)
+		if dryRun {
+			continue
+		}
+		err = svc.RabbitMQClient.PublishToLndhubExchange(context.Background(), inv, svc.EncodeInvoiceWithUserLogin)
+		if err != nil {
+			logrus.Error(err)
+		}
+	}
+	logrus.Infof("Published %d invoices", len(result))
```
bumi, line 75: should we count which ones did not have an error to be able to see if all have been published.

republish_invoices/main.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	dryRun := os.Getenv("DRY_RUN") == "true"
+	for _, inv := range result {
+		logrus.Infof("Publishing invoice with hash %s", inv.RHash)
+		if dryRun {
+			continue
+		}
+		err = svc.RabbitMQClient.PublishToLndhubExchange(context.Background(), inv, svc.EncodeInvoiceWithUserLogin)
+		if err != nil {
+			logrus.Error(err)
```
bumi, line 72: I guess this is an exception and could be a sentry notification? (though this floods sentry) - should we cancel if the publishing fails?

republish_invoices/main.go
```diff
@@ -0,0 +1,86 @@
...
+	"os"
+	"time"
+
+	"github.com/getAlby/lndhub.go/db"
+	"github.com/getAlby/lndhub.go/db/models"
+	"github.com/getAlby/lndhub.go/lib/service"
+	"github.com/getAlby/lndhub.go/rabbitmq"
+	"github.com/joho/godotenv"
+	"github.com/kelseyhightower/envconfig"
+	"github.com/sirupsen/logrus"
```
bumi, line 16: why do we use a new logger?

republish_invoices/main.go
```diff
@@ -0,0 +1,86 @@
...
+	// Open a DB connection based on the configured DATABASE_URI
+	dbConn, err := db.Open(c)
+	if err != nil {
+		log.Fatalf("Error initializing db connection: %v", err)
+	}
+	rabbitmqClient, err := rabbitmq.Dial(c.RabbitMQUri,
+		rabbitmq.WithLndInvoiceExchange(c.RabbitMQLndInvoiceExchange),
+		rabbitmq.WithLndHubInvoiceExchange(c.RabbitMQLndhubInvoiceExchange),
+		rabbitmq.WithLndInvoiceConsumerQueueName(c.RabbitMQInvoiceConsumerQueueName),
+	)
```
bumi, line 44: should this be a function like `getRabbitmqClient` to make sure we have the same client config everwhere?

bumi:
generally I would like to move towards JSON logging which helps querying log entries.

republish_invoices/main.go
kiwiidb: It's the same logger. It's what we use under the hood.

republish_invoices/main.go
```diff
@@ -50,7 +50,7 @@ func main() {
 	defer rabbitmqClient.Close()
 
 	result := []models.Invoice{}
-	err = dbConn.NewSelect().Model(&result).Where("id > ?", startId).Where("id < ?", endId).Scan(context.Background())
+	err = dbConn.NewSelect().Model(&result).Where("settled_at > ?", startDate).Where("settled_at < ?", endDate).Scan(context.Background())
```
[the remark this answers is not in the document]
kiwiidb: If `settled_at` is between those 2 values, then it's settled.

republish_invoices/main.go
bumi: how does it work? we initialize one here that we then use I think: https://github.com/getAlby/lndhub.go/blob/main/main.go#L80

republish_invoices/main.go
kiwiidb: Don't think we should cancel.

bumi (approved):

PR #355, opened 2023-05-02
Split payments

juligasa:
This PR aims to implement the splitting payments functionality inside the lndhub.
With this new functionality, a user can fetch an invoice indicating how much of the invoice and to whom is going to be split after that invoice is settled. 
One use case is multiauthor tipping when several authors contribute to a document and then that document gets tipped. The desired outcome in that case is to have an invoice that gets split in multiple once it's settled and every author gets its part. We trigger that splitting when the original invoice is paid. The other ones (sub-invoices) get automatically settled inside the lndhub accounting system.

A user can request an invoice this way `/v2/invoice?source=bafy2bzacechfogszzdyxk35nggd5udmggjmlnz5qa5oassirorr36shwssmsc&user=bahezrj4iaqacicabciqfnrov4niome6csw43r244roia35q6fiak75bmapk2zjudj3uffea,0.45&user=bahezrj4iaqacicabciqektiof7qwqp4ee3wbqew2mibgxthn7bkwr7zqvqntomi6v3bmy5q,0.48&memo=I_like_your_article&amount=1125000`
Where:
- `source=<src>` is a reference where the payment was initially generated (an article, a document section, ...)
- `user=<login>[,<pct>]` is a user login to attribute authorship. Optionally, we can specify the percentage of attribution (0-1) of this author This is optional and 1.0 is assumed if not provided. One can set multiple `user` fields
- `memo=<description>`: The regular invoice description
- `amount=<amt_msat>`: The amount in milli satoshis to be split

Then we generate as many internal invoices as authors in the request. Each internal invoice will have the total amount `amt_msat` multiplied by each author's share.  The invoice that we return to the user will be an invoice with the full amount and the payee is the house account. This account is the one collecting the payment and then distributing accordingly among authors.
Each author can view their respective invoices by calling `/v2/invoices/incoming` after they are either settled or expired/cancelled.


juligasa:
Oh, you're right!! sorry for spamming your repo!
However, I have detected minor fixes while developing that would be nice to include here. I will open the issues when merging my own branch

PR #361, opened 2023-05-07
Zerolog + user id in request log

bumi:
this uses the default `RequestLoggerWithConfig` (https://echo.labstack.com/middleware/logger/#customizable-function-based-requestlogger) and zerolog for nicer json logging. 

Having extra log tags like the user_id allows for better debugging.
So far this is mostly updating the loger syntax and adding log tags: https://github.com/getAlby/lndhub.go/pull/361/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R131-R149


rabbitmq/rabbitmq.go
```diff
@@ -108,11 +107,7 @@ func Dial(uri string, options ...ClientOption) (Client, error) {
 		consumeChannel: consumeChannel,
 		publishChannel: produceChannel,
 
-		logger: lecho.New(
-			os.Stdout,
-			lecho.WithLevel(log.DEBUG),
-			lecho.WithTimestamp(),
-		),
+		logger: zerolog.New(os.Stdout).Level(zerolog.DebugLevel).With().Timestamp().Logger(),
```
bumi, line 110: Should this be debug log level?

main.go
```diff
@@ -275,9 +291,9 @@ func main() {
 		// Setup metrics endpoint at another server
 		prom.SetMetricsPath(echoPrometheus)
 		go func() {
-			echoPrometheus.Logger = logger
-			echoPrometheus.Logger.Infof("Starting prometheus on port %d", svc.Config.PrometheusPort)
-			echoPrometheus.Logger.Fatal(echoPrometheus.Start(fmt.Sprintf(":%d", svc.Config.PrometheusPort)))
+			//echoPrometheus.Logger = logger
```
bumi, line 294: need to review and update the usage of logger here? can the zerolog logger be used?

kiwiidb (commented):
What would be the benefit of switching to zerolog exactly?
But if we're going to update the syntax of every log statement I would take the opportunity to remove the loggers from the services and use one global `log` variable.

main.go
kiwiidb: The zerolog logger can certainly be used here.

bumi:
I did it mainly becuse it was easier to customize the log tags (e.g. user_id) and it seems that this is a default echo logger. 
there is this: https://github.com/ziflex/lecho/issues/26
but in our setup with those groups it seems not possible to globally use the middleware. We would need to set the middleware on each group.

rabbitmq/rabbitmq.go
bumi: see our logs, we currently log debug logs

kiwiidb:
I feel like the diff of this PR will be too big if we do two things at once. So I would just find a way to add the user ID to the request logs using echo's `RequestLoggerWithConfig`, without changing to zerolog.

Reason is that there are many places where we log things, sometimes we use svc.Logger, sometimes we use context.logger, and if you would forget a `.Msg` to a `.Fatal` in zerolog somewhere, the functioning of the program is affected.

bumi:
This PR is actually only changing the logger. setting the user_id is only part of the log config that is anyway needed in this case: https://github.com/getAlby/lndhub.go/pull/361/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R131-R149

If we forget changing a call then it will fail, yes.. but then we simply fix it (and expand tests). :) 
Isn't also the compiler already catching those errors? 
I don't see a way of how this can be done in a smaller/incremental PR. 

but I am also OK if you know a way to get the current logger to log the user_id and that we can move the current log strings slowly to JSON logs.

kiwiidb:
> but I am also OK if you know a way to get the current logger to log the user_id and that we can move the current log strings slowly to JSON logs.

New PR: https://github.com/getAlby/lndhub.go/pull/367

PR #364, opened 2023-05-19
fix: improve rate limit handling

reneaaron:
- Fix docs
 - Remove default rate limit for non-strict endpoints
 - Change strict endpoint rate limiting to requests / second (instead of every X seconds)
 - Rate limit based on user id (if available, instead of IP) ➡️ NWC


kiwiidb (changes requested):
Looks good, I would just remove the Printf statement (if you really want to log it then you should use the logger).

PR #367, opened 2023-05-23
add user id to logs as json field

kiwiidb:
Replaces #361 

Creates the log middleware with an Enricher and injects it after the JWT middleware.
There is some code duplication because of the secured/unsecured API groups, but all in all I think it's ok.

As an example, when trying to pay a misformed invoice

Log output before (no seperate UserID field):
```
{"level":"error","id":"1ihPmGY73JFRby8AEj7ZHmN6OwiyMclO","time":"2023-05-23T19:34:23+02:00","message":"Invalid payment request user_id:26 error: rpc error: code = Unknown desc = invalid bech32 string length 3"}
{"level":"debug","id":"1ihPmGY73JFRby8AEj7ZHmN6OwiyMclO","remote_ip":"::1","host":"localhost:3000","method":"POST","uri":"/payinvoice","user_agent":"HTTPie/3.2.1","status":400,"referer":"","latency":48.579542,"latency_human":"48.579542ms","bytes_in":"18","bytes_out":"50","time":"2023-05-23T19:34:23+02:00"}
```

Log output after (`UserID` JSON field both in the request log and the error log):
```
{"level":"error","id":"qUX2veARkFzIeojX9s43uuHSSrV0Owhu","UserID":26,"time":"2023-05-23T18:20:19+02:00","message":"Invalid payment request user_id:26 error: rpc error: code = Unknown desc = invalid bech32 string length 3"}
{"level":"debug","id":"qUX2veARkFzIeojX9s43uuHSSrV0Owhu","UserID":26,"remote_ip":"::1","host":"localhost:3000","method":"POST","uri":"/payinvoice","user_agent":"HTTPie/3.2.1","status":400,"referer":"","latency":103.553458,"latency_human":"103.553458ms","bytes_in":"18","bytes_out":"50","time":"2023-05-23T18:20:19+02:00"}
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/getAlby/lndhub.go/rabbitmq"
+	"github.com/rs/zerolog"
 
 	cache "github.com/SporkHubr/echo-http-cache"
 	"github.com/SporkHubr/echo-http-cache/adapter/memory"
@@ -131,9 +132,6 @@ func main() {
 
 	e.Logger = logger
 	e.Use(middleware.RequestID())
-	e.Use(lecho.Middleware(lecho.Config{
-		Logger: logger,
-	}))
 
 	// Setup exception tracking with Sentry if configured
 	// sentry init needs to happen before the echo middlewares are added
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -307,6 +306,15 @@ func main() {
 	svc.Logger.Info("LNDhub exiting gracefully. Goodbye.")
 }
 
+func createLoggingMiddleware(logger *lecho.Logger) echo.MiddlewareFunc {
+	return lecho.Middleware(lecho.Config{
+		Logger: logger,
+		Enricher: func(c echo.Context, logger zerolog.Context) zerolog.Context {
+			return logger.Interface("UserID", c.Get("UserID"))
```
bumi, line 313: does this work if the UserID is not there or must the middleware only be added to authenticated endpoints?

main.go
```diff
@@ -186,13 +184,14 @@ func main() {
...
 	// strict rate limit for requests for sending payments
 	strictRateLimitMiddleware := createRateLimitMiddleware(c.StrictRateLimit, c.BurstRateLimit)
-	secured := e.Group("", tokens.Middleware(c.JWTSecret))
-	securedWithStrictRateLimit := e.Group("", tokens.Middleware(c.JWTSecret), strictRateLimitMiddleware)
+	secured := e.Group("", tokens.Middleware(c.JWTSecret), logMw)
+	securedWithStrictRateLimit := e.Group("", tokens.Middleware(c.JWTSecret), strictRateLimitMiddleware, logMw)
 
-	RegisterLegacyEndpoints(svc, e, secured, securedWithStrictRateLimit, strictRateLimitMiddleware, tokens.AdminTokenMiddleware(c.AdminToken))
-	RegisterV2Endpoints(svc, e, secured, securedWithStrictRateLimit, strictRateLimitMiddleware, tokens.AdminTokenMiddleware(c.AdminToken))
+	RegisterLegacyEndpoints(svc, e, secured, securedWithStrictRateLimit, strictRateLimitMiddleware, tokens.AdminTokenMiddleware(c.AdminToken), logMw)
```
bumi, line 193: do we need to pass the middleware to the individual endpoints? isn't it enough to have it on the group? (after the JWT middleware)

bumi (commented):
wahoo

main.go
kiwiidb: This is needed because these functions also register unauthenticated endpoints (/auth and /create). Otherwise it would have been enough to have it on the group.

main.go
kiwiidb: If the userID is not there it will just log `nil` . It is also added to unauthenticated endpoints, see below. We just don't add it on the landing page and on the swagger page.

main.go
kiwiidb: There is some code duplication here but that can be cleaned up later I guess.

bumi: but if the middleware works for authenticated and unauthenticated endpoints why do we need to pass it to each individually?

bumi: ok, this is that those endpoints also log at all, otherwise there would not be any logs there I guess.

PR #368, opened 2023-05-24
Feature/admin update user

kiwiidb:
This PR adds an admin endpoint for updating the user, which can be used for updating the login/password or for activating/deactivating the users account. A deactivated user can no longer fetch a new token and gets an error message when they try to do so.

The extension does not show a specific error message from the error response body, but this is what is returned in the response body:

Screenshot 2023-05-24 at 13 02 02

### DIFF
--- a/lib/service/service.go
+++ b/lib/service/service.go
@@ -3,10 +3,12 @@ package service
 import (
 	"context"
 	"fmt"
-	"github.com/getAlby/lndhub.go/rabbitmq"
 	"strconv"
 
+	"github.com/getAlby/lndhub.go/rabbitmq"
+
 	"github.com/getAlby/lndhub.go/db/models"
+	"github.com/getAlby/lndhub.go/lib/responses"
 	"github.com/getAlby/lndhub.go/lib/tokens"
 	"github.com/getAlby/lndhub.go/lnd"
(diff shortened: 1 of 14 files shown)


lib/service/service.go
```diff
@@ -57,6 +59,10 @@ func (svc *LndhubService) GenerateToken(ctx context.Context, login, password, in
 		}
 	}
 
+	if user.Deactivated {
+		return "", "", fmt.Errorf(responses.AccountDeactivatedError.Message)
+	}
+
```
bumi, line 65: can we do this check (and potential other checks in an middleware that runs before each secured endpoint. 
This has the advantage that it is checked on each request and also that this `GenerateToken` function does not have the sideeffect of checking some user details

kiwiidb: That would introduce an extra database operation for every request but it would be possible.

bumi: A lookup of the user is cheap - and I guess in many cases we do this already (though I guess we can not reuse the loaded user)

kiwiidb: We almost never do this in the request handling. We always use the user id from the token and never perform the database operation to fetch the user.

PR #370, opened 2023-05-29
Use a rabbitmq to finalize pending payments

### LINKED ISSUE #369: Update missing pending payments from rabbitmq
Currently, we track missed pending payments (initiated in an LNDhub pod that was restarted) in a goroutine at startup and do a GRPC call to LND to lookup and track the state of these payments. This has some downsides:

- With multiple LNDhub replicas we do this operation multiple times, leading to database constraint errors.
- When switching between different LND back-ends, it's possible that there are now pending payments in one lnd and if lndhub restarts with a different lnd back-end, it needs to get updates on those payments as well.
- Other back-ends like Eclair have a different API.

We propose the following system:

- At startup, LNDhub queries it's database for all pending outgoing payments (like now).
- It starts listening to RabbitMQ for updates on outgoing payments. When it finds a match (and only then), it will update the payment in the database. All other payments will already be updated in the synchronous context.
- To account for updated payments in another LNDhub pod, it will reload it's list of pending payments every hour.

LucasRouckhout:
Fixes #369


rabbitmq/rabbitmq.go
```diff
@@ -126,69 +132,125 @@ func Dial(uri string, options ...ClientOption) (Client, error) {
...
-		false,
-		// Nowait: We set this to false as we want to wait for a server response
-		// to check whether the exchange was created succesfully
-		false,
-		nil,
+func (client *DefaultClient) FinalizeInitializedPayments(ctx context.Context, svc LndHubService) error {
+	deliveryChan, err := client.amqpClient.Listen(
+		ctx,
+		client.config.lndPaymentExchange,
+		"payment.outgoing.*",
```
kiwiidb, line 141: Routing key should be a const or a config option (with this as default).

rabbitmq/rabbitmq_test.go
```diff
@@ -0,0 +1,122 @@
...
+		AnyTimes().
+		Return(models.TransactionEntry{InvoiceID: invoices[0].ID}, nil)
+
+	lndHubService.EXPECT().
+		GetTransactionEntryByInvoiceId(gomock.Any(), gomock.Eq(invoices[1].ID)).
+		AnyTimes().
+		Return(models.TransactionEntry{InvoiceID: invoices[1].ID}, nil)
+
+	ctx := context.Background()
+    successPaymnent, err := json.Marshal(&lnrpc.Payment{PaymentHash: firstHash, Status: lnrpc.Payment_SUCCEEDED})
```
kiwiidb, line 78: typo

kiwiidb (approved):

PR #387, opened 2023-06-23
Feature/lnd cluster

kiwiidb:
[Demo video](https://www.loom.com/share/c4c63c03fd764581ad4532f24878130f?sid=8900d2be-6475-4402-8fb3-6725f11220f8)

[PR overview video](https://www.loom.com/share/fffa4d124156413e95c3dd5dc933cdef)

[Comment below](https://github.com/getAlby/lndhub.go/pull/387#issuecomment-1607696929) explains why this is better than alternative solutions.

This PR introduces a cluster implementation for the LND backend type that consists of multiple LND nodes.
The cluster has an array of nodes, an active node and is doing a liveness check to ensure the active node is online and ready to make payments. The first Lightning node of the array is considered the "main" node and will always be used if it is available, the system will only switch to the 2nd node if the first node goes offline.

### DIFF
--- a/init_lnd.go
+++ b/init_lnd.go
@@ -46,10 +46,15 @@ func InitLNDCluster(c *service.Config, logger *lecho.Logger, ctx context.Context
 		}
 		nodes = append(nodes, n)
 	}
-	return &lnd.LNDCluster{
+	logger.Infof("Initialized LND cluster with %d nodes", len(nodes))
+	cluster := &lnd.LNDCluster{
 		Nodes:               nodes,
-		ActiveChannelRatio:  0.5,
+		ActiveChannelRatio:  c.LNDClusterActiveChannelRatio,
+		ActiveNode:          nodes[0],
 		Logger:              logger,
-		LivenessCheckPeriod: 30,
-	}, nil
+		LivenessCheckPeriod: c.LNDClusterLivenessPeriod,
+	}
+	//start liveness check
+	go cluster.StartLivenessLoop(ctx)
+	return cluster, nil
 }
--- a/init_lnd.go
+++ b/init_lnd.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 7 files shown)


bumi:
I am wondering if this can be done on the grpc/network level? Doesn't grpc or some kubernetes/cloud setup allow client side load balancing? 
Something like where a custom health-check is used to identify if a service is running and then traffic is routed accordingly. 
isn't this the idea of those service mesh concepts?

kiwiidb:
> I am wondering if this can be done on the grpc/network level? Doesn't grpc or some kubernetes/cloud setup allow client side load balancing? Something like where a custom health-check is used to identify if a service is running and then traffic is routed accordingly. isn't this the idea of those service mesh concepts?

I've also thought about that a lot and while this would indeed be possible I feel it would be too complex to pull off. The way to do that would mean that we would need to migrate our LND's to a k8s `Statefulset` from it's current form under a `Deployment` (Statefulset is actually more appropriate for LND). Then the traffic from LNDhub -> LND would be round-robin'ed across the 2 nodes. But:

- The way I see it we would only want to use our backup node in case the main node is down (otherwise we would need 2 equivalent nodes), and I've found no solution on how to achieve this. Maybe @LucasRouckhout knows about how to do non-round-robin load balancing in this case?
- We would need to write a custom LivenessCheck script that does the same as this code: it needs to call `GetInfo`, it needs to parse the response and check the nr. of active channels in order to know if the node is ready to start making payments.
- We would need to copy over the `admin.macaroon` between the 2 nodes, and we would need to make sure the TLS certificate is valid for the shared host.
- We need to know if an invoice is internal or external. If we abstract this all away there is no way of knowing what the public keys of the nodes that are in our cluster are.

So in light of these arguments I find the approach of writing a new client implementation better.

bumi:
Aren't those things like linkerd only proxies for normal deployments? 

for me this sounds like some default traffic routing and I am concerned about the complexity this requires within this application. 
did you look into things like haproxy?

PR #391, opened 2023-06-30
reserve fee amount in database

kiwiidb:
This PR adds the maximum fee a user can pay to the amount that is debited from their account.
In case of a succesful payment, this provisional max amount is credited to their balance again.
In case of a failed payment, the transaction entry is reversed anyway, but I had to change this to use the right amount (the one from the entry to reverse, which was the same as the invoice amount up until now, but will be different now).

To do:

- [ ] Add integration test
- [ ] Make demo video about issue & fix


kiwiidb:
Replaced by #395

PR #393, opened 2023-07-03
update dependencies

kiwiidb:
In order to update the datadog dependency I had to update all dependencies otherwise I ran into conflicts.


bumi:
yay update all the things :D

PR #407, opened 2023-08-10
chore: add dependabot

reneaaron:


kiwiidb:
we already have dependabot?
https://github.com/getAlby/lndhub.go/blob/main/.github/dependabot.yaml

reneaaron:
I just looked at the configuration in the GitHub settings (there it was disabled) and didn't check the repository. 🙈

PR #413, opened 2023-09-05
feat: add created_at to create invoice response

rolznz:
untested

### DIFF
--- a/controllers_v2/invoice.ctrl.go
+++ b/controllers_v2/invoice.ctrl.go
@@ -134,6 +134,7 @@ type AddInvoiceResponseBody struct {
 	PaymentHash    string    `json:"payment_hash"`
 	PaymentRequest string    `json:"payment_request"`
 	ExpiresAt      time.Time `json:"expires_at"`
+	CreatedAt      time.Time `json:"created_at"`
 }
 
 // AddInvoice godoc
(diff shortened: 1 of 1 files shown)


kiwiidb (approved):

controllers_v2/invoice.ctrl.go
```diff
@@ -134,6 +134,7 @@ type AddInvoiceResponseBody struct {
 	PaymentHash    string    `json:"payment_hash"`
 	PaymentRequest string    `json:"payment_request"`
 	ExpiresAt      time.Time `json:"expires_at"`
+	CreatedAt      time.Time `json:"created_at"`
```
bumi, line 137: can we return this as timestamp? otherwise we have to deal with timezones?

controllers_v2/invoice.ctrl.go
```diff
@@ -174,6 +175,7 @@ func (controller *InvoiceController) AddInvoice(c echo.Context) error {
 		PaymentHash:    invoice.RHash,
 		PaymentRequest: invoice.PaymentRequest,
 		ExpiresAt:      invoice.ExpiresAt.Time,
+		CreatedAt:      invoice.CreatedAt,
```
bumi, line 178: why is it `CreatedAt` here but `ExpiresAt.Time` above?

rolznz: expiresAt is nullable, you can create an invoice that does not expire

controllers_v2/invoice.ctrl.go
rolznz: We cannot change expires_at - it needs to be backward compatible. I think created_at should be consistent here.

We could add 2 new fields for timestamp format. What do you think? @bumi @kiwiidb

rolznz: CC @jankoegel

PR #414, opened 2023-09-07
Chore: RabbitMQ logging

### LINKED ISSUE #411: Add JSON logging to rabbitmq publishing
Currently we log something like: 
```
Successfully published invoice to rabbitmq with RHash  xxxx
```

this makes it harder to search for it in the logs. We should use JSON logging and add metadata of the invoice.

kiwiidb:
Fixes #411 
This PR adds metadata fields to the logs from the processes that use Rabbitmq: payment consumer, invoice consumer and publisher routines. This will help us identify issues and also allow us to trace payments/invoices across services.

I also propose a standardized way of naming metadata fields (I would propose snake case), so we can search for them in datadag across different service logs:
 - `payment_hash` (only one used in this PR)
 - `lndhub_user_id`, `getalbycom_user_id`
 - `invoice_id`
 - `message`
 - `error` 
 - etc

We would need some follow up PR's in order to implement this kind of structured logging across LNDhub & other services.

### DIFF
--- a/rabbitmq/rabbitmq.go
+++ b/rabbitmq/rabbitmq.go
@@ -205,6 +205,11 @@ func (client *DefaultClient) FinalizeInitializedPayments(ctx context.Context, sv
 
 					continue
 				}
+				client.logger.Infoj(log.JSON{
+					"subroutine":   "payment finalizer",
+					"payment_hash": invoice.RHash,
+					"message":      "updating payment",
+				})
 
 				switch payment.Status {
 				case lnrpc.Payment_SUCCEEDED:
@@ -222,7 +227,11 @@ func (client *DefaultClient) FinalizeInitializedPayments(ctx context.Context, sv
 						continue
 					}
 
-					client.logger.Infof("Payment finalizer: updated successful payment with hash: %s", payment.PaymentHash)
+					client.logger.Infoj(log.JSON{
+						"subroutine":   "payment finalizer",
+						"message":      "updated succesful payment",
+						"payment_hash": payment.PaymentHash,
+					})
 					delete(pendingInvoices, payment.PaymentHash)
 
 				case lnrpc.Payment_FAILED:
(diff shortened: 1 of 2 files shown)


rabbitmq/rabbitmq.go
```diff
@@ -352,16 +419,20 @@ func (client *DefaultClient) publishToLndhubExchange(ctx context.Context, invoic
...
+		"payment_hash": invoice.RHash,
+	})
 
 	return nil
 }
 
-func captureErr(logger *lecho.Logger, err error) {
-	logger.Error(err)
+func captureErr(logger *lecho.Logger, err error, j log.JSON) {
+	j["error"] = err
```
bumi, line 435: what is this adding to the logging? the error message?

rabbitmq/rabbitmq.go
```diff
@@ -352,16 +419,20 @@ func (client *DefaultClient) publishToLndhubExchange(ctx context.Context, invoic
...
+	logger.Errorj(j)
 	sentry.CaptureException(err)
```
bumi, line 437: when I currently look at the sentry stack traces it does not give much information and it does not contain the original stack trace but links to this `CaptureException` error

I read here something about a different erros package and wrapping an error to keep the stack trace? 
https://incident.io/blog/golang-errors
does that make sense?

rabbitmq/rabbitmq.go
```diff
@@ -352,16 +419,20 @@ func (client *DefaultClient) publishToLndhubExchange(ctx context.Context, invoic
...
-		captureErr(client.logger, err)
 		return err
 	}
 
-	client.logger.Debugf("Successfully published invoice to rabbitmq with RHash %s", invoice.RHash)
+	client.logger.Infoj(log.JSON{
+		"subroutine":   "invoice publisher",
+		"message":      "succesfully published invoice",
+		"payment_hash": invoice.RHash,
+	})
```
bumi, line 429: is some more metadata helpful like the key (and exchange?) 
the key is computed it might be interesting to add.

bumi (commented):
good idea with the subroutine field!

rabbitmq/rabbitmq.go
kiwiidb: Go stacktraces inherently require more work, yes. So that would be something to add later.

PR #429, opened 2023-09-20
chore: add missing logs

### LINKED ISSUE #419: log unlogged errors
For example, [here](https://github.com/getAlby/lndhub.go/blob/56b569b49dcaef04f98e9173e1256546687edb3d/controllers_v2/invoice.ctrl.go#L202).

Find all instances where we don't log the error.

im-adithya:
Fixes #419

### DIFF
--- a/controllers/addinvoice.ctrl.go
+++ b/controllers/addinvoice.ctrl.go
@@ -50,6 +50,7 @@ func AddInvoice(c echo.Context, svc *service.LndhubService, userID int64) error 
 
 	amount, err := svc.ParseInt(body.Amount)
 	if err != nil || amount < 0 {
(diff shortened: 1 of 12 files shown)


controllers/addinvoice.ctrl.go
```diff
@@ -50,6 +50,7 @@ func AddInvoice(c echo.Context, svc *service.LndhubService, userID int64) error
 
 	amount, err := svc.ParseInt(body.Amount)
 	if err != nil || amount < 0 {
+		c.Logger().Errorf("Invalid amount %v for user_id:%v error: %v", amount, userID, err)
```
bumi, line 53: we should use JSON logging

bumi (commented):
make sure to log as JSON

kiwiidb (approved):

PR #430, opened 2023-09-20
chore: add rhash and preimage to invoice before failure

### LINKED ISSUE #418: Keysend: add r_hash, destination_custom_records
We don't add these at the moment so we can't identify the payments if they fail

im-adithya:
Fixes #418


im-adithya:
Didn't add `r_hash` yet

kiwiidb (changes requested):
Looks good but I really would like to have a test that reproduces the error like it used to be.

kiwiidb (approved):
I've checked the test on the main branch and it can reproduce the issue.
tACK

PR #432, opened 2023-09-25
Feature/add max volume

kiwiidb:
This PR implements an extra check when doing payments. If the total transaction volume of the account (incoming + outgoing) over a certain period is bigger than the configured values, it will block the payment (and send a sentry notification). 

- Fix existing tests (were broken due to issues introduced with #392 )
- Refactor `BalanceCheck` to `CheckPaymentAllowed` which will now perform the balance check and the volume check
- Add extra error response type for volume exceeded
- integration test

Default configuration is no limit.

example response:

```
http POST localhost:3000/v2/payments/keysend < keysend.json Authorization:"Bearer $token";
HTTP/1.1 400 Bad Request
Content-Length: 112
Content-Type: application/json; charset=UTF-8
Date: Mon, 25 Sep 2023 08:52:50 GMT
X-Request-Id: M3bd0Hkrvg0e7qg0e5rZX9glAUeWUEKE

{
    "code": 2,
    "error": true,
    "message": "transaction volume too high. please contact support for further assistance."
}
```

### DIFF
--- a/lib/service/user.go
+++ b/lib/service/user.go
@@ -208,7 +208,6 @@ func (svc *LndhubService) GetVolumeOverPeriod(ctx context.Context, userId int64,
 	err = svc.DB.NewSelect().Table("invoices").
 		ColumnExpr("sum(invoices.amount) as result").
 		Where("invoices.user_id = ?", userId).
-		Where("invoices.state = ?", common.InvoiceStateSettled).
 		Where("invoices.settled_at >= ?", time.Now().Add(-1*period)).
 		Scan(ctx, &result)
 	if err != nil {
--- a/lib/service/user.go
+++ b/lib/service/user.go
@@ -133,7 +133,7 @@ func (svc *LndhubService) CheckPaymentAllowed(ctx context.Context, lnpayReq *lnd
 	if svc.Config.FeeReserve {
 		minimumBalance += svc.CalcFeeLimit(lnpayReq.PayReq.Destination, lnpayReq.PayReq.NumSatoshis)
 	}
-	if currentBalance >= minimumBalance {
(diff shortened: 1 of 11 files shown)


lib/service/user.go
```diff
@@ -185,3 +199,17 @@ func (svc *LndhubService) InvoicesFor(ctx context.Context, userId int64, invoice
 	}
 	return invoices, nil
 }
+
+func (svc *LndhubService) GetVolumeOverPeriod(ctx context.Context, userId int64, period time.Duration) (result int64, err error) {
+
+	err = svc.DB.NewSelect().Table("invoices").
```
bumi, line 205: todo: check if we have the correct index for this query

lib/service/user.go
```diff
@@ -121,17 +124,28 @@ func (svc *LndhubService) FindUserByLogin(ctx context.Context, login string) (*m
...
-	return currentBalance >= minimumBalance, nil
+	if currentBalance < minimumBalance {
+		return &responses.NotEnoughBalanceError, nil
+	}
+	volume, err := svc.GetVolumeOverPeriod(ctx, userId, time.Duration(svc.Config.MaxVolumePeriod*int64(time.Second)))
+	if err != nil {
+		return nil, err
+	}
+	if volume > svc.Config.MaxVolume {
+		sentry.CaptureMessage(fmt.Sprintf("transaction volume exceeded for user %d", userId))
```
bumi, line 145: do we need a sentry notification there? 🤔 or is a log enough? 
but either way I think a log would be good.

kiwiidb: These error messages are logged in the level above.

PR #437, opened 2023-10-09
legacy endpoint: no errored invoices

kiwiidb:
As discussed on [slack](https://getalby.slack.com/archives/C02TDCWNX0T/p1696837586971999)

This stops failed payments from showing up in Zeus.

### DIFF
--- a/controllers/gettxs.ctrl.go
+++ b/controllers/gettxs.ctrl.go
@@ -65,6 +65,10 @@ func (controller *GetTXSController) GetTXS(c echo.Context) error {
 
 	response := make([]OutgoingInvoice, len(invoices))
 	for i, invoice := range invoices {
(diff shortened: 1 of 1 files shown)


bumi:
should we filter this in the `controller.svc.InvoicesFor` function? if we know we don't want those we could also query the ones with the settled filter.

kiwiidb:
That will also impact the response of the incoming invoice endpoint (then we no longer show unpaid created invoices).

bumi:
> That will also impact the response of the incoming invoice endpoint (then we no longer show unpaid created invoices).

we would need to pass the state that we want to get to the `controller.svc.InvoicesFor` function.

bumi:
let's ship this 🚢

PR #442, opened 2023-10-30
chore: migrate internal and keysend columns to be non null

### LINKED ISSUE #417: Default values for internal / keysend
For boolean fields like `internal` or `keysend` we currently do not have a default value. 
This means that we set it to `true` but otherwise it's NULL.

for consistence we should have a default value of false there. (easiest done with a DB migration)

im-adithya:
Fixes #417


bumi:
I think we would need to set a default? or how does the value now get fille?

db/models/invoice.go
```diff
@@ -22,8 +22,8 @@ type Invoice struct {
 	DestinationCustomRecords map[uint64][]byte `json:"custom_records,omitempty"`
 	RHash                    string            `json:"r_hash"`
 	Preimage                 string            `json:"preimage" bun:",nullzero"`
-	Internal                 bool              `json:"-" bun:",nullzero"`
-	Keysend                  bool              `json:"keysend" bun:",nullzero"`
+	Internal                 bool              `json:"-" bun:",notnull,default:false"`
+	Keysend                  bool              `json:"keysend" bun:",notnull,default:false"`
```
bumi, line 26: they still use nullzero:
https://bun.uptrace.dev/guide/models.html#default

`nullzero,notnull,default:false` ?

I think if `""` (blank) is set then bun converts it to `NULL` which then uses the default of `false`.

db/migrations/20231030140000_invoice_keysend.up.sql
```diff
@@ -0,0 +1,9 @@
+UPDATE invoices
+SET keysend = false
+WHERE keysend IS NULL;
+
+ALTER TABLE invoices
+ALTER COLUMN keysend SET DEFAULT false;
+
+ALTER TABLE invoices
+ALTER COLUMN keysend SET NOT NULL;
```
bumi, line 9: if a default is set, do we then still need a `NOT NULL` constraint?

im-adithya: Yes because I've seen the db values go from false to NULL while running integration_tests for some reason

PR #453, opened 2023-11-23
feat: include amount in index

reneaaron:
Speed up the `check_balance()` trigger that is being run after every payment.


bumi:
@reneaaron see the test error. how did we do it with other migrations?

PR #454, opened 2023-11-24
chore: optimize keysend/multikeysend balance checks

im-adithya:
## Description

Moves `CheckPaymentAllowed` check one level up so that it doesn't add much load by triggering everytime during multikeysend checks

### DIFF
--- a/controllers_v2/keysend.ctrl.go
+++ b/controllers_v2/keysend.ctrl.go
@@ -81,7 +81,22 @@ func (controller *KeySendController) KeySend(c echo.Context) error {
 		c.Logger().Errorf("Invalid keysend request body: %v", err)
 		return c.JSON(http.StatusBadRequest, responses.BadArgumentsError)
 	}
-
+	syntheticPayReq := &lnd.LNPayReq{
+		PayReq: &lnrpc.PayReq{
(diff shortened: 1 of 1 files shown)


controllers_v2/keysend.ctrl.go
```diff
@@ -149,6 +160,26 @@ func (controller *KeySendController) MultiKeySend(c echo.Context) error {
...
+		},
+		Keysend: true,
+	}
+	resp, err := controller.svc.CheckPaymentAllowed(ctx, syntheticPayReq, userID)
+	if resp != nil {
+		controller.svc.Logger.Errorf("User does not have enough balance user_id:%v amount:%v", userID, syntheticPayReq.PayReq.NumSatoshis)
+		return resp
+	}
+	if err != nil {
+		errResp := responses.GeneralServerError
```
kiwiidb, line 176: Check the error first, and log the error and just return a GeneralServerError here.

kiwiidb (changes requested):

controllers_v2/keysend.ctrl.go
im-adithya: Done!

kiwiidb (approved):

PR #459, opened 2023-11-29
chore: add max balance check while creating invoice

im-adithya:
Adds max balance check in v2 controller

### DIFF
--- a/lib/service/user.go
+++ b/lib/service/user.go
@@ -165,6 +165,19 @@ func (svc *LndhubService) CalcFeeLimit(destination string, amount int64) int64 {
 	return limit
 }
 
+func (svc *LndhubService) CheckUserBalance(ctx context.Context, amount, userId int64) (result *responses.ErrorResponse, err error) {
+	if svc.Config.MaxAccountBalance > 0 {
+		currentBalance, err := svc.CurrentUserBalance(ctx, userId)
+		if err != nil {
+			return nil, err
+		}
+		if currentBalance+amount > svc.Config.MaxAccountBalance {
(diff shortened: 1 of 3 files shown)


lib/service/user.go
```diff
@@ -165,6 +165,19 @@ func (svc *LndhubService) CalcFeeLimit(destination string, amount int64) int64 {
...
 }
 
+func (svc *LndhubService) CheckUserBalance(ctx context.Context, amount, userId int64) (result *responses.ErrorResponse, err error) {
+	if svc.Config.MaxAccountBalance > 0 {
+		currentBalance, err := svc.CurrentUserBalance(ctx, userId)
+		if err != nil {
+			return nil, err
+		}
+		if currentBalance+amount > svc.Config.MaxAccountBalance {
+			return &responses.BadArgumentsError, fmt.Errorf("Max account balance exceeded for user_id:%d (balance:%d + amount:%d)", userId, currentBalance, amount)
```
bumi, line 175: is this the best error to return here? 
or do we return a generic error that we then use in the controller?

lib/service/user.go
```diff
@@ -165,6 +165,19 @@ func (svc *LndhubService) CalcFeeLimit(destination string, amount int64) int64 {
 	return limit
 }
 
+func (svc *LndhubService) CheckUserBalance(ctx context.Context, amount, userId int64) (result *responses.ErrorResponse, err error) {
+	if svc.Config.MaxAccountBalance > 0 {
+		currentBalance, err := svc.CurrentUserBalance(ctx, userId)
+		if err != nil {
+			return nil, err
+		}
+		if currentBalance+amount > svc.Config.MaxAccountBalance {
```
bumi, line 174: @kiwiidb so far in the other check we only check the settled existing payments and don't include the amount?

what do you think is best here?

PR #465, opened 2023-12-05
feat: fetch and use the limits from the JWT Token

### LINKED ISSUE #460: Allow the volume configuration to be defined in the auth JWT
Currently the configurations are globally defined.
Is it possible to define those on a request level using data from the JWT?

im-adithya:
Fixes #460

This would enable setting different limits for different users instead of just one global setting.

⚠️ TODO: Add tests

### DIFF
--- a/controllers/addinvoice.ctrl.go
+++ b/controllers/addinvoice.ctrl.go
@@ -37,7 +37,7 @@ func (controller *AddInvoiceController) AddInvoice(c echo.Context) error {
 
 func AddInvoice(c echo.Context, svc *service.LndhubService, userID int64) error {
 	var body AddInvoiceRequestBody
-	limits := svc.GetLimitsFromContext(c)
+	limits := svc.GetLimits(c)
 
 	if err := c.Bind(&body); err != nil {
 		c.Logger().Errorf("Failed to load addinvoice request body: %v", err)
--- a/controllers/addinvoice.ctrl.go
+++ b/controllers/addinvoice.ctrl.go
@@ -37,6 +37,7 @@ func (controller *AddInvoiceController) AddInvoice(c echo.Context) error {
 
 func AddInvoice(c echo.Context, svc *service.LndhubService, userID int64) error {
 	var body AddInvoiceRequestBody
+	limits := svc.GetLimitsFromContext(c)
 
 	if err := c.Bind(&body); err != nil {
 		c.Logger().Errorf("Failed to load addinvoice request body: %v", err)
@@ -61,7 +62,7 @@ func AddInvoice(c echo.Context, svc *service.LndhubService, userID int64) error 
 		return c.JSON(http.StatusBadRequest, responses.BadArgumentsError)
 	}
 
-	resp, err := svc.CheckIncomingPaymentAllowed(c.Request().Context(), amount, userID)
+	resp, err := svc.CheckIncomingPaymentAllowed(c.Request().Context(), amount, userID, limits)
 	if err != nil {
 		return c.JSON(http.StatusInternalServerError, responses.GeneralServerError)
 	}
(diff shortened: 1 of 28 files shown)


lnd/config.go
```diff
@@ -22,6 +22,14 @@ type Config struct {
 	LNDClusterPubkeys            string  `envconfig:"LND_CLUSTER_PUBKEYS"` //comma-seperated list of public keys of the cluster
 }
 
+type Limits struct {
```
kiwiidb, line 25: These limits are unrelated to LND and thus this struct should not be in the LND package.

lib/tokens/jwt.go
```diff
@@ -15,12 +16,17 @@ import (
...
+	MaxSendVolume      int64 `json:"maxSendVolume"`
+	MaxSendAmount      int64 `json:"maxSendAmount"`
+	MaxReceiveVolume   int64 `json:"maxReceiveVolume"`
+	MaxReceiveAmount   int64 `json:"maxReceiveAmount"`
+	MaxAccountBalance  int64 `json:"maxAccountBalance"`
 	jwt.StandardClaims
 }
 
-func Middleware(secret []byte) echo.MiddlewareFunc {
+func Middleware(secret []byte, limits *lnd.Limits) echo.MiddlewareFunc {
```
kiwiidb, line 29: I would change nothing to the signature of this function.

lnd/config.go
kiwiidb, line 25: These limits are unrelated to LND and thus this struct should not be in the LND package.

lib/tokens/jwt.go
kiwiidb, line 29: I would not change the arguments of this function.

lib/tokens/jwt.go
```diff
@@ -38,6 +44,31 @@ func Middleware(secret []byte) echo.MiddlewareFunc {
 		token := c.Get("UserJwt").(*jwt.Token)
 		claims := token.Claims.(*jwtCustomClaims)
 		c.Set("UserID", claims.ID)
+		if limits.MaxSendVolume == 0 {
```
kiwiidb, line 47: I would just call C.set for everything, no if, no else.

lib/service/user.go
```diff
@@ -279,3 +274,24 @@ func (svc *LndhubService) GetVolumeOverPeriod(ctx context.Context, userId int64,
 	}
 	return result, nil
 }
+
+func (svc *LndhubService) GetLimitsFromContext(c echo.Context) (limits *lnd.Limits) {
```
kiwiidb, line 278: I would create a function defined on `svc *LNDhubService` with the default limits, that are fetched from `svc.Config`. Then I would overwrite these values with the values from the context, if these are larger than 0 (since they will be set always they will never be nil)

controllers/addinvoice.ctrl.go
```diff
@@ -37,6 +37,7 @@ func (controller *AddInvoiceController) AddInvoice(c echo.Context) error {
 
 func AddInvoice(c echo.Context, svc *service.LndhubService, userID int64) error {
 	var body AddInvoiceRequestBody
+	limits := svc.GetLimitsFromContext(c)
```
kiwiidb, line 40: There is no need to already do `GetLimitsFromContext` here. We have the context all the way down, we can just fetch the limits when we need them, there is no need to change the arguments of all these functions

kiwiidb (changes requested):
I think the PR can be simplified quite a bit, see comments.

lnd/config.go
im-adithya: Shifted to `db/models/user.go`

lib/service/user.go
im-adithya: Changed to `GetLimits`

kiwiidb (approved):

PR #467, opened 2023-12-09
fix: exclude erroneous payments

### LINKED ISSUE #466: Exclude errorneous payments from APIs
When a payment fails (due to e.g. a timeout) a new invoice is created and it's state is set to `Error`. 

API endpoints should not return these kind of transactions as they don't contain any information, anyway:

image

The filtering could take place over here: 

https://github.com/getAlby/lndhub.go/blob/feat/add-transaction-entries-index/lib/service/user.go#L196

reneaaron:
Fixes #466

### DIFF
--- a/integration_tests/internal_payment_test.go
+++ b/integration_tests/internal_payment_test.go
@@ -394,7 +394,7 @@ func (suite *PaymentTestSuite) TestInternalPaymentFail() {
 	_ = suite.createPayInvoiceReqError(bobInvoice.PayReq, suite.aliceToken)
 
 	userId := getUserIdFromToken(suite.aliceToken)
-	invoices, err := suite.service.InvoicesFor(context.Background(), userId, common.InvoiceTypeOutgoing)
+	invoices, err := invoicesFor(suite.service, userId, common.InvoiceTypeOutgoing)
 	if err != nil {
 		fmt.Printf("Error when getting invoices %v\n", err.Error())
 	}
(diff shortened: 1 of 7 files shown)


integration_tests/internal_payment_test.go
```diff
@@ -399,10 +399,9 @@ func (suite *PaymentTestSuite) TestInternalPaymentFail() {
 		fmt.Printf("Error when getting invoices %v\n", err.Error())
 	}
 
-	// check if first one is settled, but second one error (they are ordered desc by id)
-	assert.Equal(suite.T(), 2, len(invoices))
-	assert.Equal(suite.T(), common.InvoiceStateError, invoices[0].State)
-	assert.Equal(suite.T(), common.InvoiceStateSettled, invoices[1].State)
+	// check if first one is settled, second one is erroneous and hence isn't listed
+	assert.Equal(suite.T(), 1, len(invoices))
+	assert.Equal(suite.T(), common.InvoiceStateSettled, invoices[0].State)
```
bumi, line 404: can we query for the second one differently and check that it has the state error?

bumi (commented):
can we load the other invoice differently and still check that it is there and has the error state?

integration_tests/internal_payment_test.go
im-adithya: Fixed

bumi (approved):

reneaaron:
Follow-up: #469

PR #470, opened 2023-12-18
small updates to errors.go

itstomekk:
fixed error texts


bumi:
thanks for the PR. but actually we can not do this here in the open source project. 
@reneaaron what do you think if we add a new "code" and then show this error where this API is used?

reneaaron:
Sure, that would mean integrating those error codes in the js-sdk, right?

PR #475, opened 2023-12-30
fix: migrating payment failed errors to json logging

reneaaron:


controllers/keysend.ctrl.go
```diff
@@ -108,7 +108,15 @@ func (controller *KeySendController) KeySend(c echo.Context) error {
 	}
 	sendPaymentResponse, err := controller.svc.PayInvoice(c.Request().Context(), invoice)
 	if err != nil {
-		c.Logger().Errorf("Payment failed: user_id:%v error: %v", userID, err)
+		c.Logger().Errorj(
+			log.JSON{
+				"message": 	"payment failed",
+				"error": err,
+				"lndhub_user_id": userID,
```
bumi, line 115: in the app here we use `user_id` so far. 
should we move to specifying it as `lndhub_user_id` or is the app + `user_id` combination enough?

PR #484, opened 2024-02-21
Feature/reconciliation invoices

frnandu:


bumi:
Something must be wrong here, because this is what we already have in here: https://github.com/getAlby/lndhub.go/blob/main/cmd/payment-reconciliation/main.go
isn't it?

bumi:
this merge looks wrong: https://github.com/getAlby/lndhub.go/pull/484/commits/adf8cde1cf0b00b118116cd749af448edb223e66

I assume that file of the feature branch should have a different name. - it probably was based on the file in the master branch but then changed but the name stayed the same. 

when you merged master in the branch you used the code from master then. 

so we should move that in a file with a different name.

PR #499, opened 2026-06-12
Replace lnrpc.SendPaymentSync with routerrpc.SendPaymentV2

### LINKED ISSUE #498: SendPaymentSync was removed in lnd 0.21
The program can no longer make payments because the API method SendPaymentSync [was removed](https://github.com/lightningnetwork/lnd/blob/master/docs/release-notes/release-notes-0.21.0.md#breaking-changes) in the latest stable lnd. Ironically, I switched to lndhub.go from the original lndhub because of a similar issue, SendPayment method was removed too so lndhub stopped working. I thought lndhub.go switched to routerrpc already but apparently it did not.

rkfg:
Fixes #498. lnd 0.21.0-beta removed multiple RPC methods and as such it became impossible to make payments with lndhub.go (still possible to receive). This patch switches payments to `routerrpc.SendPaymentV2`.

The change is relatively big, some fields in SendPaymentV2 request/response are different from the old API, amounts can be specified in both sats and msats (mutually exclusively) and so on. I had to upgrade the Go version since lnd requires Go 1.25 at least. There are some changes in tests as well as minor fixes in formatting. `google.golang.org/protobuf` needs to be overridden according to https://github.com/lightningnetwork/lnd/issues/8326#issuecomment-1873794540


raucao:
Considering the amount of vulnerabilities Red Team has disovered over the last week or so, I think being able to upgrade LND should probably be the number one priority right now.

rkfg:
True. I've been running this patch since the day I published it, no issues whatsoever. Though none are expected, the change is pretty trivial. It's one of the veeeery few changes I did fully manually these days without AI, whether that's good or bad is up to everyone to decide ﻿😂﻿

raucao:
Good to know, thanks. I'm having an agent run a PR review, will have it post its findings here.

raucao:
Couldn't get Copilot to reply itself, so here's what it found:

---

## PR Review Findings (Functionality + Security)

### 1) High severity: `SendPaymentV2` stream is handled as if it were a single final response
**File:** `/home/runner/work/lndhub.go/lndhub.go/lib/service/invoices.go`  
**Function:** `SendPaymentSync`

The new implementation calls `svc.LndClient.SendPaymentSync(...)`, then does a single `Recv()` and treats that one message as final success/failure.

`routerrpc.SendPaymentV2` is a stream of status updates; the first message is not guaranteed to be terminal. The code should process updates until a terminal state (success/failure) is reached.

**Impact:** false payment failures (or other wrong status handling), triggering retries and potentially duplicate actual payment attempts. This is a financial-integrity risk.

---

### 2) High severity: payment hash/preimage binary encoding regression for legacy response format
**File:** `/home/runner/work/lndhub.go/lndhub.go/lib/service/invoices.go`  
**Function:** `SendPaymentSync`

The new code sets:
- `sendPaymentResponse.PaymentPreimage = []byte(preimage)`
- `sendPaymentResponse.PaymentHash = []byte(paymentHash)`

In this flow, `preimage` and `paymentHash` are string values. Converting with `[]byte(...)` yields ASCII/UTF-8 bytes of the hex text, not decoded raw bytes.

Legacy endpoints return these fields in `JavaScriptBuffer` form and expect raw bytes. This breaks compatibility and can produce incorrect payloads for clients expecting binary values.

**Impact:** wrong `payment_hash` / `payment_preimage` bytes in legacy API responses.

---

### 3) Medium severity: amountless BOLT11 invoice payments likely broken
**File:** `/home/runner/work/lndhub.go/lndhub.go/lib/service/invoices.go`  
**Function:** `createLnRpcSendRequest`

For non-keysend payments, the migrated `routerrpc.SendPaymentRequest` sets `PaymentRequest` and `FeeLimitSat`, but does not set explicit amount.

Amountless invoices require caller-supplied amount. In the previous flow amount was set on the send request; now this appears omitted.

**Impact:** valid amountless invoices may fail unexpectedly.

---

## Security summary

- No direct auth-bypass, injection, or secret-leak issue identified in the changed business logic itself.
- The most important security-relevant risk is **#1**: incorrect streaming payment-state handling can cause bad retry behavior and financial loss scenarios.

raucao:
@rkfg Did you see the code review?

rkfg:
Yeah, didn't have the time to fix it yet. Feel free to adopt the PR if in a hurry!

raucao:
Thanks for the update.

Personally, I'm not experienced enough in Go to make judgement calls on changesets like this. Hopefully, one of the maintainers can have a look soon.

raucao:
Maybe this inspires a sense of urgency:

CLN users just got notified that they should take their node offline until the next release, or upgrade immediately via precompiled emergency binaries: https://stacker.news/items/1555439?commentId=1555585

